PR #5, opened 2019-11-15
Changed HTML-serving endpoints to only serve HTML

JakeUrban:
Reason:
I discovered django rendering issues from hitting Polaris endpoints while testing from stellar-anchor-server.

Problem:
For our `/interactive_deposit`, `/interactive_withdraw`, and `/more_info` endpoints, we were returning an HTML form for a successful request and JSON for a failed request. This was causing django to fail while attempting to render the response.

It turns out that Django Rest Framework doesn't play nicely with returning different content types from the same endpoint, see [the docs](https://www.django-rest-framework.org/api-guide/renderers/). Essentially, the caller of the endpoint must specify the content type they're expecting, otherwise Django Rest Framework expects our endpoints to return the default content type for the function.

Solution:
I converted each of these endpoints to only return HTML, even on failure. From the caller's perspective, it would have been confusing to expect HTML and receive JSON anyway. The HTML rendered on failure is a new and simple `error.html` template that displays the error message and status code.

Extas:
I also added a `settings.py` file explicitly for testing, since some settings needed to be changed to run locally.

### DIFF
--- a/polaris/polaris/deposit/views.py
+++ b/polaris/polaris/deposit/views.py
@@ -16,8 +16,9 @@ from django.urls import reverse
 from django.views.decorators.clickjacking import xframe_options_exempt
 from django.core.management import call_command
 from rest_framework import status
-from rest_framework.decorators import api_view
+from rest_framework.decorators import api_view, renderer_classes
 from rest_framework.response import Response
+from rest_framework.renderers import TemplateHTMLRenderer
 from stellar_sdk.keypair import Keypair
 from stellar_sdk.exceptions import Ed25519PublicKeyInvalidError
 
@@ -137,6 +138,7 @@ def confirm_transaction(request):
 
 @xframe_options_exempt
 @api_view(["GET", "POST"])
+@renderer_classes([TemplateHTMLRenderer])
 def interactive_deposit(request):
     """
     `GET /deposit/interactive_deposit` opens a form used to input information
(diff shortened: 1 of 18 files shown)


polaris/polaris/deposit/views.py
```diff
@@ -146,15 +148,15 @@ def interactive_deposit(request):
     # Validate query parameters: account, asset_code, transaction_id.
     account = request.GET.get("account")
     if not account:
-        return render_error_response("no 'account' provided")
+        return render_error_response("no 'account' provided", content_type="text/html")
```
msfeldstein: is there a way to do this at a higher level rather than putting content_type="text/html" everywhere?

polaris/polaris/fee/views.py
```diff
@@ -21,8 +21,8 @@ def _op_type_is_valid(asset_code: str, operation: str, op_type: str) -> bool:
     return False
 
 
-@validate_sep10_token()
 @api_view()
+@validate_sep10_token()
```
msfeldstein: Does the order of these matter?

polaris/polaris/helpers.py
```diff
@@ -28,10 +28,22 @@ def calc_fee(asset: Asset, operation: str, amount: float) -> float:
...
+    Renders an error response in Django.
+
+    Currently supports HTML or JSON responses.
+    """
+    resp_data = {
+        "data": {"error": description, "status_code": status_code},
+        "status": status_code,
+        "content_type": content_type
+    }
+    if content_type == "text/html":
```
msfeldstein: Should this function take the request, and determine the content type based on the header of the request?  That way applications can say whether they want html or json instead of us choosing based on endpoint

polaris/polaris/fee/views.py
JakeUrban: Unfortunately yes. I tested and if `@validate_sep10_token` was above `@api_view`, the endpoint wouldn't be able to render properly.

polaris/polaris/helpers.py
JakeUrban: From a code perspective I see why we wouldn't want to have to add `content_type="text/html"` for each of these calls, but I don't think allowing the caller to decide what content type to receive is appropriate here.

To me, it doesn't make sense for our endpoints that return JSON to optionally return HTML (what would it be?), and in the same vein, our `/interactive_*` endpoints must return an interactive form, not JSON. So since having multiple content types per endpoint doesn't really serve our use cases, I'd vote to keep the approach we have in this branch currently.

An alternative solution would be to create a new function for HTML-returning error responses, instead of using the same function for JSON and HTML errors.

polaris/polaris/deposit/views.py
JakeUrban: Answered in your final comment below

PR #6, opened 2019-11-18
Convert to new POST endpoints

msfeldstein:
In stellar/stellar-protocol#438 we change the /deposit and /withdraw endpoints to new URLs and change them to POST instead of GET for security. This makes the changes specified in that PR.

Changes /deposit to /transactions/deposit/interactive
Changes /withdraw to /transactions/withdraw/interactive
Changes both endpoints to POST

Land with stellar/sep24-demo-client#133


JakeUrban:
These changes will be released with version 0.9.2 or 1.0, depending on the urgency of getting these changes deployed

PR #7, opened 2019-11-19
0.9.2.dev to master

JakeUrban:
Only one change is included in 0.9.2: https://github.com/stellar/django-polaris/pull/6

I want to review this change again. Here are a few things I'm seeing:
- The SEP-24 protocol documentation and doc strings (less important) aren't up to date. Particularly, `/deposit` and `/withdraw` are no longer listed in the API Endpoints section but they are still referenced throughout the document. This ties into my next point.
- This change moves `/deposit` to `/transactions/deposit/interactive` but also keeps `/deposit/confirm_transaction` and `/deposit/interactive_deposit` where they are
    - The SEP and PR appear to communicate that `/deposit` and `/withdraw` were completely removed. I personally feel like we should either move all `/deposit` and `/withdraw` routes under `/transactions`, or none at all. Right now, its hard to understand the reasoning behind why we moved `/deposit` if we were planning to keep the other endpoints the same.
    - It is no longer clear what the difference is between `/transactions/deposit/interactive` and `deposit/interactive_deposit`. I think having `/transactions/deposit` and `/transactions/deposit/interactive` makes more sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,10 @@ At the time of writing this, the latest public release is v0.9.
 After each release, a new development branch is created for the next upcoming release. In this case, that is 0.9.1.dev.
 Follow the instructions below to submit a valid pull request.
 
-1. Checkout the current development branch (0.9.1.dev) (`git checkout`)
-1. Create a new branch based off the latest dev branch (`git checkout -b`)
-1. Make your changes (`git add .; git commit -m`)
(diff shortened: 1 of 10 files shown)


msfeldstein:
I think we can rename /deposit/interactive_deposit to /transactions/deposit/webapp or something.  I used `/transactions/deposit/interactive` instead of `/transactions/deposit` because there will also be a non-interactive /deposit endpoint.

Issue #8, opened 2019-11-21
SEP10 JWT contains Server address and not Client address


leighmcculloch:
### What version are you using?

`v0.9.2`

### What did you do?

I haven't used polaris yet but I took a look at the SEP-10 implementation here:
https://github.com/stellar/django-polaris/blob/1da3dd3ffd2e59dfbc19e27aceaeb097b5ad94f3/polaris/polaris/sep10auth/views.py#L99-L117

### What did you expect to see?

I expected to see the address/public-key of the Client's Stellar account set as the `sub` field of the JWT. The Client's address is the source account of the challenge transaction's first manage data operation.

According to [SEP-10's definition of what is in a token] the `sub` field should contain:
>the public key of the authenticating Stellar account (G...)

The `authenticating` Stellar account is the client.

### What did you see instead?

The address/public-key of the distribution Stellar account set as the `sub` field of the JWT.

[SEP-10's definition of what is in a token]: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#token

JakeUrban:
Thanks Leigh,

This should do the trick:
```
def _generate_jwt(request, envelope_xdr):
    """
    Generates the JSON web token from the challenge transaction XDR.

    See: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0010.md#token
    """
    issued_at = time.time()
    transaction_envelope = TransactionEnvelope.from_xdr(
        envelope_xdr,
        network_passphrase=settings.STELLAR_NETWORK_PASSPHRASE
    )
    transaction = transaction_envelope.transaction
    hash_hex = binascii.hexlify(transaction_envelope.hash()).decode()
    jwt_dict = {
        "iss": request.build_absolute_uri("/auth"),
        "sub": transaction.source.public_key,
        "iat": issued_at,
        "exp": issued_at + 24 * 60 * 60,
        "jti": hash_hex,
    }
    encoded_jwt = jwt.encode(jwt_dict, settings.SERVER_JWT_KEY, algorithm="HS256")
    return encoded_jwt.decode("ascii")
```

JakeUrban:
Turns out the transaction's source is the anchor's account since its a challenge generated by the server. So instead of assigning `sub` to `transaction.source.public_key`, we'll use `transaction.operations[0].source` as defined in SEP-10:

* `transaction`: an XDR-encoded Stellar transaction with the following:
  * source account set to server's signing account
[...]

PR #13, opened 2019-11-23
v0.9.3 Changes

JakeUrban:
I will document the changes here but I encourage you to check out the HTML docs.
```
git clone https://github.com/JakeUrban/django-polaris.git
cd django-polaris
pipenv install --dev
cd docs
pipenv run make html
open build/html/index.html
```

Changes:
Documentation
- I used Sphinx and the autodoc extension which reads the code's docstrings to populate the documentation

The `integrations` module
- Contains `DepositIntegration`, `WithdrawalIntegration`, and `register_integrations`
- See the code and docs for info on these

Docker & Django Commands
- Added a new django command `poll_pending_deposits`
- `poll_pending_deposits` is a substitute for `/confirm_transaction`. I didn't remove the endpoint but I can if we decide its no longer needed. We'll have to update the demo client to no longer make the confirmation API call if we want to remove it.

JWT authentication
- Added  to `/more_info` and `/confirm_transaction`
- Fixed issue #8 (replaced `sub` field of JWT with client account)

Deposit:
- Changed 403 response to 200 on `/transactions/deposit/interactive`

Transactions:
- Added `status_message` column to the `Transaction` DB model. This is an internal field to store messages associated with the status, such as when a transaction was unable to be executed on the stellar network.


polaris/polaris/deposit/utils.py
```diff
@@ -0,0 +1,139 @@
...
+        source_account=server_account,
+        network_passphrase=settings.STELLAR_NETWORK_PASSPHRASE,
+        base_fee=base_fee
+    )
+    try:
+        server.load_account(stellar_account)
+    except BaseHorizonError as address_exc:
+        # 404 code corresponds to Resource Missing.
+        if address_exc.status != 404:
+            transaction.status_message = (
```
msfeldstein: do we need to update the `transaction.status` to `error` here as well as `status_message`

polaris/polaris/deposit/utils.py
```diff
@@ -0,0 +1,139 @@
...
+        transaction_envelope = builder.append_create_account_op(
+            destination=stellar_account,
+            starting_balance=starting_balance,
+            source=settings.STELLAR_DISTRIBUTION_ACCOUNT_ADDRESS
+        ).build()
+        transaction_envelope.sign(settings.STELLAR_DISTRIBUTION_ACCOUNT_SEED)
+        try:
+            server.submit_transaction(transaction_envelope)
+        except BaseHorizonError as submit_exc:
+            transaction.status_message = (
```
msfeldstein: error status here as well

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -63,51 +62,100 @@ def process_withdrawal(response, transaction):
...
+    the status as `pending_stellar`. If exception was raised during
+    `process_withdrawal`, we mark the corresponding `Transaction` status as
+    `error` and save an error message to the transaction database object.
+    If the Stellar transaction succeeded, we mark it as `completed`.
+
+    :param error_msg: a description of the error that has occurred.
+    :param response: a response body returned from Horizon for the transaction
+    :param transaction: a database model object representing the transaction
+    """
+    if not response["successful"]:
```
msfeldstein: What is the difference between `not successful` and having an error?  wouldn't someting with an error_msg also be not successful?

polaris/polaris/models.py
```diff
@@ -99,6 +101,9 @@ class Transaction(models.Model):
         choices=MEMO_TYPES, default=MEMO_TYPES.text, max_length=10
     )
 
+    objects = models.Manager()
```
msfeldstein: what does this do?

polaris/polaris/transaction/views.py
```diff
@@ -74,13 +75,14 @@ def _construct_more_info_url(request):
 @xframe_options_exempt
 @api_view()
 @renderer_classes([TemplateHTMLRenderer])
-def more_info(request):
+@validate_sep10_token(content_type="text/html")
```
msfeldstein: How do we expect a JWT header to be sent to the more_info url?  It's just a url that's opened in a popup, often not something we can send headers to.

JakeUrban: Good catch, I'll remove it. However this endpoint should be protected somehow, since its showing transaction details. Maybe we use that one-time-token strategy here as well.

polaris/polaris/models.py
JakeUrban: We already use it throughout the codebase, `Transaction.objects.filter(...)`. It's a builtin attribute for all django models, used for forming queries. I just defined it in the class so PyCharm would stop complaining.

polaris/polaris/management/commands/watch_transactions.py
JakeUrban: I don't particularly like the logic here either, I think its confusing, but I wanted to maintain our current functionality. Previously, if `response["successful"]` was `False`, we would save the `Transaction.status` as `pending_stellar`. Note that we now save `error_msg` to `Transaction.status_message` in this case.

As @debnil explained it, that status means the transaction was submitted to the stellar network but not "confirmed" as per [SEP-24](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#fields-for-withdraw-transactions). If we choose to start assigning unsuccessful transactions the `error` status, there also isn't a distinction between Stellar network failures and anchor server errors.

Errors originating from processing the transaction after it was submitted should use the `error` status. Specifically, errors raised in `process_withdrawal()`.

I'm totally open to suggestions on how this should work. Maybe we should think harder about error handling and reporting outside of this PR, though. Right now we don't provide good support for debugging issues related to failed transactions.

polaris/polaris/deposit/utils.py
JakeUrban: That is what I had originally had implemented, but I realized that was technically the wrong status to use. See my comment explaining my thought process on `error` vs. `pending_stellar`.

README.rst
```diff
@@ -0,0 +1,113 @@
...
+
+    pip install pipenv
+    git clone https://github.com/stellar/django-polaris.git
+    cd django-polaris
+    pipenv install --dev
+
+Git Flow
+--------
+At the time of writing this, the latest public release is v0.9.
+After each release, a new development branch is created for the next upcoming release. In this case, that is 0.9.1.dev.
```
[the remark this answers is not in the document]
JakeUrban: Yea I agree, I'll remove it.

polaris/polaris/integrations/transactions.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @classmethod
+    def after_deposit(cls, transaction: Transaction):
+        """
+        Use this function to perform any post-processing of `transaction` after
+        its been executed on the Stellar network. This could include actions
+        such as updating other Django models in your project or logging
+        relevant information. Overriding this function is not required
+
+        :param transaction: The django Transaction object that was executed on
```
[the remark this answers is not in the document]
JakeUrban: We can if we want to add documentation for the models defined in Polaris. I began doing this originally but realized that most fields were self explanatory, but for columns such as `status`, it would probably be useful.

polaris/polaris/integrations/transactions.py
```diff
@@ -0,0 +1,99 @@
...
+            "`poll_transactions` must be implemented to process deposits"
+        )
+
+    @classmethod
+    def after_deposit(cls, transaction: Transaction):
+        """
+        Use this function to perform any post-processing of `transaction` after
+        its been executed on the Stellar network. This could include actions
+        such as updating other Django models in your project or logging
+        relevant information. Overriding this function is not required
```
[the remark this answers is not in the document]
JakeUrban: Sure, good example.

JakeUrban:
Updated docs at: https://html.jakeurban.now.sh/

Issue #22, opened 2019-12-02
Transaction records should be created upon first POST


msfeldstein:
When a client POSTs to `/transaction/withdraw/interactive` or `/transaction/deposit/interactive` it should create a persisted Transaction object to the db.  Currently we create an ephemeral UUID, and then upon submission of the interactive webapp we use that UUID to actually create a Transaction object.  The problem here is that since there's no record, the `/transactions` endpoint is empty so it looks like no transaction has been started.  It's even more confusing since the transaction originally returned does have an ID so it seems like we should be able to query it, but can't.

JakeUrban:
This will be released as a part of the Form Integrations functionality #27

JakeUrban:
Released with 0.9.8

Issue #19, opened 2019-11-26
Redirect to /more_info after successful deposit


JakeUrban:
When you submit the form from `/interactive_deposit` it succeeds and renders the `more_info.html` template, but the url is still `/interactive_deposit`.  This makes it so you can't refresh the page without trying to re-submit the form and it errors out.  

Instead, we should submit the form, then redirect to the more info url since we're done with `interactive_deposit`

JakeUrban:
Released with 0.9.8

PR #27, opened 2019-12-03
Form Integrations & KYC Support

JakeUrban:
[Docs](https://html.jakeurban.now.sh)
[Design](https://docs.google.com/document/d/1R_Jzt8-TkBfWsrMRwfKaLpaoW-hhaCqm6eS1NkftEqY/edit#)

- Introduces the `TransactionForm` base class
- Added a couple functions to `DepositIntegration` and `WithdrawalIntegration`
    - `form_for_transaction()`
    - `after_form_validation()`
- Adjusted /webapp endpoints to use the new integrations
- Fixes Issues #22 and #19


JakeUrban:
@msfeldstein hold off on the review -- the KYC design exposed a flaw in the form integrations design. Once thats adjusted you can review.

JakeUrban:
Ok @msfeldstein This is ready for a review

JakeUrban:
We've decided to redesign parts of this feature, which will cover KYC use cases. The review can wait until this redesign is pushed.

docs/integrations/index.rst
```diff
@@ -1,24 +1,84 @@
...
 .. _stellar-anchor-server: https://github.com/stellar/stellar-anchor-server
 
-Polaris does most of the work implementing SEP-24_. However, Polaris simply
-doesn't have the information it needs to interface with an anchor's partner
-financial entities. This is where :class:`.DepositIntegration` and
-:class:`.WithdrawalIntegration` come in.
+Polaris does most of the work implementing SEP-24_. However, some pieces of
+SEP-24 can only be implemented by the anchor. Specifically, anchors need to
+implement their own banking rails and KYC requirements. This is where
+:class:`DepositIntegration` and :class:`WithdrawalIntegration` come in.
```
msfeldstein: "This is where the Integrations classes come in" since KYC is added now and Deposit/WithdrawIntegration doesn't cover all the integrations

docs/integrations/index.rst
```diff
@@ -1,24 +1,84 @@
...
+
+.. _Bulma: https://bulma.io/documentation
+.. _Django Forms: https://docs.djangoproject.com/en/2.2/topics/forms/#forms-in-django
+
+Polaris uses `Django Forms`_ for collecting users' information, such as their
+email or how much of their asset they want to deposit. Polaris comes out of
+the box with forms for deposit and withdrawal flows.
+
+However, the data collected may not be sufficient for your needs. Maybe you
+need to do collect additional fields and some validation or processing with
```
msfeldstein: s/Maybe you need to do collect/Maybe you need to collect

polaris/polaris/deposit/views.py
```diff
@@ -141,50 +139,69 @@ def confirm_transaction(account: str, request: Request) -> Response:
...
 
-    form = DepositForm(request.POST)
-    form.asset = asset
-    # If the form is valid, we create a transaction pending external action
-    # and render the success page.
+    # request.method == "POST"
+    form = rdi.form_for_transaction(transaction)(request.POST)
+    is_transaction_form = issubclass(form.__class__, TransactionForm)
+    if is_transaction_form:
+        form.asset = asset
```
msfeldstein: Why do we need the asset to live in the form? It doesn't seem like we'd want to show this as a form element to the user, its not something they can change

polaris/polaris/deposit/views.py
```diff
@@ -141,50 +139,69 @@ def confirm_transaction(account: str, request: Request) -> Response:
...
+            )
+            transaction.save()
+
+        # Perform any defined post-validation logic defined by Polaris users
+        rdi.after_form_validation(form, transaction)
+        # Check to see if there is another form to render
+        form_class = rdi.form_for_transaction(transaction)
+
+        if form_class:
+            return Response(
```
msfeldstein: I get the feeling the GET and POST methods should be different endpoints, both for code clarity, but also in case the page needs to reload.  
For example `POST /webapp/submit_form` will process the request then redirect to `GET /webapp` to show the next available form.  Otherwise, if the page needs to reload for any reason the browser/webview will want to rePOST the previous form.

I think this is not a big deal for now but we may want to change this in the future

polaris/polaris/integrations/transactions.py
```diff
@@ -19,8 +26,8 @@ def poll_pending_deposits(cls, pending_deposits: QuerySet
         **OVERRIDE REQUIRED**
 
         This function should poll the financial entity for the state of all
-        `pending_deposits` transactions and return the ones ready to be
-        executed on the Stellar network.
+        `pending_deposits` and return the ones ready to be executed on the
```
msfeldstein: Clarify that it returns the ones where the external deposit has been completed.  "ready to be executed on the stellar network" implies that the external deposit has been completed but it couldn't help to be explicit

polaris/polaris/integrations/transactions.py
```diff
@@ -45,6 +53,7 @@ def poll_pending_deposits(cls, pending_deposits: QuerySet
         memory usage.
 
         :param pending_deposits: a django Queryset for pending Transactions
+        :return: a list of Transaction database objects to execute
```
msfeldstein: `a list of Transaction database objects to execute` feels confusing, it sounds like its going to execute a database transaction, it should be explicit that its Transaction entries for transactions which have successful user deposits and are ready to be executed on the stellar network

polaris/polaris/templates/deposit/form.html
```diff
@@ -16,13 +16,18 @@
 
     {% for field in form %}
     <div class="field">
+      {% if field.name == 'amount' %}
```
msfeldstein: Is fa-dollar-sign a bulma attribute? If so, can we pass it in in the `django.Form` instead of special casing amount here?

polaris/polaris/deposit/views.py
JakeUrban: `form.asset` is used in the validation function for the _amount_ field, which is called from `form.is_valid()`. We don't control the parameters for the validation function so we have to give the form access to the non-field object somehow.

`asset` is not shown in the form, only attributes of type `Field` are displayed.

polaris/polaris/deposit/views.py
JakeUrban: Thats fair, I'll make a separate PR for it. #34

polaris/polaris/templates/deposit/form.html
JakeUrban: I don't think so because the `<span>` tag and its contents are a separate component from the form field. The `fa-dollar-sign` attribute probably needs to live in an `<i>` tag too.

PR #31, opened 2019-12-06
Fixed issues #22 and #19

JakeUrban:
See the issues for a detailed description.

Essentially, we moved Transaction creation to `POST /transactions/<transaction>/interactive` from `POST /transactions/<transaction>/webapp`, and changed the response from `POST /transactions/<transaction>/interactive` to a redirect to `more_info`, instead of returning the `more_info` template from the same endpoint.


polaris/polaris/deposit/views.py
```diff
@@ -238,6 +222,14 @@ def deposit(account: str, request: Request) -> Response:
 
     # Construct interactive deposit pop-up URL.
     transaction_id = create_transaction_id()
+    Transaction.objects.create(
+        id=transaction_id,
+        stellar_account=account,
+        asset=asset,
+        kind=Transaction.KIND.deposit,
+        status=Transaction.STATUS.pending_user_transfer_start,
```
msfeldstein: this status should be `incomplete` at this point right?

polaris/polaris/withdraw/views.py
```diff
@@ -147,7 +118,25 @@ def withdraw(account: str, request: Request) -> Response:
...
+    # appropriately sized string for the `HashMemo`.
     transaction_id = create_transaction_id()
+    transaction_id_hex = transaction_id.hex
+    withdraw_memo = "0" * (64 - len(transaction_id_hex)) + transaction_id_hex
+    Transaction.objects.create(
+        id=transaction_id,
+        stellar_account=account,
+        asset=asset,
+        kind=Transaction.KIND.withdrawal,
+        status=Transaction.STATUS.pending_user_transfer_start,
```
msfeldstein: same here

Issue #38, opened 2019-12-09
Feature Request: Authentication for Interactive Flows


JakeUrban:
Implementation ticket for #17 
[Design](https://docs.google.com/document/d/15upOcP2ZLtGui97oCZGe3lwnTwATAcpdUEASsQh5K2M/edit#)

JakeUrban:
Found a bug after merging to master. PR for fix: #43

I'll close this ticket after its been merged to master.

PR #39, opened 2019-12-10
Interactive Authentication

JakeUrban:
Issue: #38 
[Design](https://docs.google.com/document/d/15upOcP2ZLtGui97oCZGe3lwnTwATAcpdUEASsQh5K2M/edit#)

Implements our authentication design for interactive flows.


polaris/polaris/helpers.py
```diff
@@ -120,3 +126,95 @@ def validate_jwt_request(request: Request) -> str:
...
+def validate_interactive_jwt(content_type: str = "text/html") -> Callable:
+    """
+    Decorator to validate the JWT generated by /interactive endpoints
+    for /webapp endpoints.
+    """
+
+    def decorator(view) -> Callable:
+        def wrapper(request: Request, *args, **kwargs) -> Response:
+            try:
+                if request.method == "GET":
```
msfeldstein: It seems odd to determine the action by the request method.  Would it make sense to have a separate decorator for the first case `authenticate_interactive_jwt` that converts the jwt to a session, and then `authenticate_interactive_session` for actually checking the session authentication?  It would be clearer in the code (`validate_interactive_jwt` isn't actually whats happening for either case) and would also allow us to not have a problem if we need to have any GET requests in the flow.

polaris/polaris/deposit/views.py
```diff
@@ -83,18 +86,15 @@ def _verify_optional_args(request):
 @xframe_options_exempt
 @api_view(["GET", "POST"])
 @renderer_classes([TemplateHTMLRenderer])
+@validate_interactive_jwt()
```
msfeldstein: It seems like someone could validate their own account, but then request info on a Transaction that they don't own.  As long as there is a session authenticated, they could request any transaction ID in the function below.

polaris/polaris/helpers.py
JakeUrban: I like that solution, the only issue is that we'd have 2 decorators on the same function, and only 1 of them would take affect, while the other one would have to simply pass the view function along.

If we pair this change with your suggestion to split up the `GET` and `POST` requests into separate endpoints, having two decorators would make a lot more sense.

polaris/polaris/deposit/views.py
JakeUrban: Thats true. One idea is to not only save `request.session["authenticated"]` but also `request.session["account"]`. Then in `POST` requests, check that the requested transaction belongs to the account on the request.

That seems to solve this issue, but let me know if you have a better solution.

polaris/polaris/helpers.py
JakeUrban: I renamed the decorator to `@interactive_authentication` and filed an issue to split the interactive endpoints by method.

polaris/polaris/deposit/views.py
msfeldstein: yeah i think that's exactly right, session should store some user identifier

Issue #51, opened 2019-12-14
Feature Request: Integration for toml fields


JakeUrban:
Polaris needs to provide an integration point for anchors to specify the correct information for their toml file.

PR #62, opened 2019-12-17
Defined toml_func integration point, updated docs

JakeUrban:
#51

### DIFF
--- a/example/server/apps.py
+++ b/example/server/apps.py
@@ -7,11 +7,9 @@ class AnchorConfig(AppConfig):
     def ready(self):
         from polaris.integrations import register_integrations
         from .integrations import (MyDepositIntegration,
-                                   MyWithdrawalIntegration,
-                                   get_stellar_toml)
+                                   MyWithdrawalIntegration)
 
         register_integrations(
             deposit=MyDepositIntegration(),
-            withdrawal=MyWithdrawalIntegration(),
-            toml_func=get_stellar_toml
+            withdrawal=MyWithdrawalIntegration()
         )
--- a/example/server/apps.py
+++ b/example/server/apps.py
@@ -6,8 +6,12 @@ class AnchorConfig(AppConfig):
 
     def ready(self):
         from polaris.integrations import register_integrations
-        from .integrations import MyDepositIntegration, MyWithdrawalIntegration
+        from .integrations import (MyDepositIntegration,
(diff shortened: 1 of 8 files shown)


JakeUrban:
CircleCI failed because I updated the `example` code for the new changes, but `example` is tested by installing the `django-polaris` PyPI package listed in its Pipfile, which doesn't have the changes.

We have a two options here:
- have `example` install polaris from the local source when testing in CircleCI
    - The problem with this is that `example` may work for the local source but not the `django-polaris` package it will install in production, and we wouldn't know
- Don't update `example` code until the relevant `polaris` code has been officially released on PyPI.

JakeUrban:
I'll remove the updates to `example` to fix the failing test cases

msfeldstein:
can we run the `example` test cases using the dev setup to point to the actual code?  This can be a follow up pr but would be good to have the full test suite run before landing polaris changes

example/server/apps.py
```diff
@@ -6,8 +6,10 @@ class AnchorConfig(AppConfig):
 
     def ready(self):
         from polaris.integrations import register_integrations
-        from .integrations import MyDepositIntegration, MyWithdrawalIntegration
+        from .integrations import (MyDepositIntegration,
```
msfeldstein: I've noticed a lot of formatting changes in the PR, can we do a PR thats just running the formatter on all the files, with no other changes, so we don't have unrelated changes in other PRs?

polaris/polaris/integrations/forms.py
```diff
@@ -26,6 +26,10 @@ class TransactionForm(forms.Form):
 
     def clean_amount(self):
         """Validate the provided amount of an asset."""
+        # TODO: do we want all amounts to be rounded?
```
msfeldstein: Lets keep PRs focused on one change.

But i don't think we should round 1.123USD to 1.12USD, you should be able to trade fractions of a cent.

example/server/apps.py
JakeUrban: Sure. I just added a pre-commit hook for this repo to use `black` formatting.

polaris/polaris/integrations/forms.py
JakeUrban: Do banks allow that? What if someone enters 1.123 but can only transfer 1.12 through their bank?

Issue #42, opened 2019-12-10
Document developer instructions


msfeldstein:
It's currently difficult to develop (outside of unit tests) and we should make it easier.  The current solution isn't documented, but involves running a local PyPI server, and seems kind of complicated to do with docker.

We need to document how to get up and running and provide as smooth as possible of an experience to develop polaris features.

PR #63, opened 2019-12-17
Updated docs and README for contributing

JakeUrban:
#42


overcat:
Are the contents of **README.rst** and **src/index.rst** identical? Maybe we don't need to write the same content in **src/index.rst**. Just use `.. include:: ../README.rst` to include it.

Refence: https://stackoverflow.com/questions/44563794/how-to-correctly-include-other-rest-files-in-a-sphinx-project

msfeldstein:
what overcat said.  If that's difficult no problem

PR #66, opened 2019-12-18
Template Integrations

JakeUrban:
[Design](https://docs.google.com/document/d/1ozpkq_J65bnwMavg7_RkUyYsmTnTd6YOYXn8GB-4uPE/edit?folder=1UGE_vSPMeJP0Ml3RsSwTiehX_9sloZ9y#heading=h.jst5k0gzlzpd)

All Polaris templates, and the static assets linked to those templates, are now replaceable. This includes:
- `deposit/form.html`
- `withdraw/form.html`
- `transaction/more_info.html`
- `error.html`

`base.html` is never served on its own so there is no need to provide functionality to override it. Additionally, all templates and static assets, including the ones listed above, have been renamed with a `polairs` prefix to help the anchor avoid naming conflicts.


JakeUrban:
After speaking with the team we've decided to table this PR and implement a design that gives anchors less control. If anchors want to fully control the frontend stack for the interactive flow, they can do that by providing a URL for an endpoint other than `/webapp`.

Issue #55, opened 2019-12-16
Reference Server: Mock Banking Rails


JakeUrban:
We should create a mock interface for fake banking rails and have the reference server use it to provide an example for others.

PR #69, opened 2019-12-20
Reference Implementation: Mock Banking Rails

JakeUrban:
#55 

Implements a mock banking rails interface and uses it in the `example` app server. This should provide more clarity for how to use the Polaris integration interface.

### DIFF
--- a/example/manage.py
+++ b/example/manage.py
@@ -6,7 +6,7 @@ import sys
 
 def main():
     """Main function for executing arguments."""
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "settings")
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "server.settings")
     try:
         from django.core.management import execute_from_command_line
     except ImportError as exc:
--- a/example/server/integrations.py
+++ b/example/server/integrations.py
@@ -30,10 +30,10 @@ class MyDepositIntegration(DepositIntegration):
 
         # interface with mock banking rails
         ready_deposits = []
-        rails_client = rails.RailsClient(settings.MOCK_BANK_ACCOUNT_ID)
+        client = rails.BankAPIClient(settings.MOCK_BANK_ACCOUNT_ID)
         for deposit in pending_deposits:
-            rails_deposit = rails_client.get_deposit(memo=deposit.external_extra)
-            if rails_deposit and rails_deposit.status == "complete":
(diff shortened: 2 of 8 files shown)


example/server/integrations.py
```diff
@@ -18,29 +22,83 @@ def poll_pending_deposits(cls, pending_deposits: QuerySet) -> List[Transaction]:
         Anchors should implement their banking rails here, as described
         in the :class:`.DepositIntegration` docstrings.
 
-        For the purposes of this reference implementation, we simply return
-        all pending deposits.
+        This implementation interfaces with a fake banking rails client
+        for demonstration purposes.
         """
+        # act like we're doing more work than we are for demo purposes
```
msfeldstein: Let's put this in the RailsClient, where the actual work would be done.  Perhaps say this is mimicking doing a fake API network request for your bank.

example/server/mock_banking_rails.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(self, to_account: RailsAccount, amount: Decimal, memo: str):
+        self.id = str(uuid4())
+        self.to_account = to_account
+        self.from_account = RailsAccount(str(uuid4()))
+        self.amount = amount
+        self.status = "complete"
+        self.memo = memo
+
+
+class RailsClient:
```
msfeldstein: Can we call this MyBankAdapter or something? RailsClient doesn't seem very clear, it should be an adapter between polaris and the banking api right?

example/manage.py
```diff
@@ -6,7 +6,7 @@
 
 def main():
     """Main function for executing arguments."""
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "settings")
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "server.settings")
```
msfeldstein: What are these changes?

example/server/integrations.py
msfeldstein: actually i see why its out here, because we make N requests to `client.get_deposit`.  nabd.

example/manage.py
JakeUrban: I moved the settings file under `server` so I had to adjust the python path to it.

example/server/mock_banking_rails.py
JakeUrban: Sure

PR #73, opened 2020-01-07
Implement external interactive flow integration

JakeUrban:
Adds the integration point needed to allow anchors to use an external application for the interactive flow.

### DIFF
--- a/docs/integrations/index.rst
+++ b/docs/integrations/index.rst
@@ -4,7 +4,6 @@ Integrations
 
 .. _SEP-24: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md
 .. _Django Commands: https://docs.djangoproject.com/en/2.2/howto/custom-management-commands
-.. _stellar-anchor-server: https://github.com/stellar/stellar-anchor-server
 
 Polaris does most of the work implementing SEP-24_. However, some pieces of
 SEP-24 can only be implemented by the anchor. Specifically, anchors need to
@@ -56,11 +55,6 @@ See the :doc:`../forms/index` documentation for the `TransactionForm` definition
 
 .. autofunction:: polaris.integrations.WithdrawalIntegration.after_form_validation
 
-Implementation Example
-^^^^^^^^^^^^^^^^^^^^^^
-The stellar-anchor-server_ implements many of these functions outlined above. For
-reference examples, see the github repository.
-
 Form CSS
 ^^^^^^^^
 Polaris uses default CSS provided by Bulma_ for styling forms. To keep the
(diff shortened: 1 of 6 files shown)


docs/integrations/index.rst
```diff
@@ -75,6 +69,26 @@ The `attrs` parameter adds a HTML attribute to the `<input>` tag that Bulma
 uses to add better styling. You may also add more Bulma-supported attributes
 to Polaris forms.
 
+Use a External Application for the Interactive Flow
```
JakeUrban: an

docs/integrations/index.rst
```diff
@@ -75,6 +69,26 @@ The `attrs` parameter adds a HTML attribute to the `<input>` tag that Bulma
...
+
+.. _described in SEP-24: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#2-interactive-customer-information-needed
+
+Polaris provides `Form Integrations`_ for collecting and processing information
+about a deposit or withdraw. However if you would rather use another
+application to collect that information, override this function to return the
+URL that should be requested to begin that process.
+
+Note that if you choose to use another application or set of endpoints,
+you must make a ``postMessage`` call back to the window that opened the
```
msfeldstein: postMessage or the URL callback approach

msfeldstein: Instead of having the external application handle the callbacks which can be complicated, would it make sense to have them just redirect to a polaris `/interactive/complete` url which handles the callbacks properly?

msfeldstein:
looks good but lets handle the completion state for them

JakeUrban:
> looks good but lets handle the completion state for them

I think the best way to do this is just have them request the `/transaction/more_info` page, which has the postMessage call we use by default.

msfeldstein:
That's something that just happens to work right now, but its not the cleanest thing.  I think the best thing to do is redirect to a /completion endpoint, and on our side we can just 302 redirect that to the more_info_url.  This way if we change the behavior of the more_info page, we don't have to tell the anchors to change their implementations, we can simply have the /completion endpoint do something different.

docs/integrations/index.rst
```diff
@@ -80,11 +80,12 @@ application to collect that information, override this function to return the
...
-transaction. This signals to the wallet that the anchor is done processing the
-transaction and may resume control. If you don't want to end by displaying the
-more info page, you must make a ``postMessage`` call to the window that opened
-the interactive flow popup or support a callback argument as `described in SEP-24`_.
+you must redirect to the `/transactions/<deposit or withdraw>/interactive/complete?id=`
+endpoint for the relevant transaction when finished. This signals to the wallet that
+the anchor is done processing the transaction and may resume control. If you
+don't want to end by displaying the more info page, you must make a
+``postMessage`` call to the window that opened the interactive flow popup or
+support a callback argument as `described in SEP-24`_.
```
msfeldstein: > If you
more info page, you must make a ``postMessage`` call to the window that opened	don't want to end by displaying the more info page, you must make a
the interactive flow popup or support a callback argument as `described in SEP-24`_.	``postMessage`` call 

I wouldn't even mention this.  If someone wants it we can add guidance but lets not advertise that this is possible.

JakeUrban: Ok, sounds good.

Issue #54, opened 2019-12-16
Reference Server: Add KYC collection and User tracking


JakeUrban:
We should add a KYC form to our reference server and start storing the user info in our DB. We want other anchors to do this so users do not have to repeatedly submit their KYC info for every transaction.

PR #74, opened 2020-01-08
Added user tracking and KYC form

JakeUrban:
#54 

Added three models to the `server` app in `/example`:
- `PolarisUser`: a user interfacing with a polaris-support transfer server
- `PolarisStellarAccount`: The stellar accounts this anchor has processed transactions for
- `PolarisUserTransaction`: A two-column table to map `PolarisStellarAccount`s to `Transaction` objects

Added a `KYCForm` to the Form Integrations flow to collect contact info.
- This form is only served if we don't have a database entry for the stellar account passed

Together, these models and form are used to keep a record of each user's activity. This should serve as an example to developers using Polaris.


example/server/integrations.py
```diff
@@ -1,20 +1,79 @@
...
+def track_user_activity(form: forms.Form, transaction: Transaction):
+    """
+    Creates a PolarisUserTransaction object, and depending on the form
+    passed, also creates a new PolarisUserAccount and potentially a
+    new PolarisUser. This function ensures an accurate record of a
+    particular person's activity.
+    """
+    if isinstance(form, KYCForm):
+        data = form.cleaned_data
+        user = PolarisUser.objects.filter(email=data.get("email")).first()
```
msfeldstein: Does this mean i can create transactions in anyones account just by putting in their email?

example/server/models.py
```diff
@@ -0,0 +1,34 @@
...
+    email = models.EmailField(unique=True)
+
+    objects = models.Manager()
+
+    @property
+    def name(self):
+        return " ".join([str(self.first_name), str(self.last_name)])
+
+
+class PolarisUserAccount(models.Model):
```
msfeldstein: This should be PolarisStellarAccount since it represents a stellar account, not an anchors user account

example/server/models.py
```diff
@@ -0,0 +1,34 @@
...
+    user = models.ForeignKey(PolarisUser, on_delete=models.CASCADE)
+    account = models.CharField(max_length=254, unique=True)
```
msfeldstein: should we set an exact length = 56 here since its always going to be a stellar public key just for some extra data integrity validation?

example/server/integrations.py
JakeUrban: Yes.. thats true. What would a normal anchor do? I would guess they'd send a confirmation email and wait to continue processing until the user confirms right?

msfeldstein: Maybe the best thing is to just not deal with emails, and link every user account to a single stellar account?  I don't really want to design a separate user login system at this point

JakeUrban: I'm not sure that would really be a useful example. I don't think implementing the confirmation email would be difficult, and the user would only have to do it once. From thereon out, we'll get the PolarisUser from the account the wallet passes us.

This situation also demonstrates that anchor's may want to serve pages without forms during the interactive flow, like when waiting on the user to confirm their email. I filed an issue for this: #78

msfeldstein: cool i would also be ok with adding a todo and issue so you can land this

Issue #33, opened 2019-12-06
Internationalization


msfeldstein:
We need a way to internationalize any human readable information coming from polaris.  The most obvious is the Form UI in the interactive flow, but also includes things like error messages from api requests and asset information.

We should implement an internationalization solution, localize to one other language, and create documentation on how anchors should localize their own content.

PR #76, opened 2020-01-09
Internationalization

JakeUrban:
#33 

Adds full backend support for internationalization, specifically Portuguese. Anchors wanting other languages can make a pull request containing the .po files and adjusted `LANGUAGES` setting.

Changes:
- Any text rendered to the user was wrapped with `_()` (a django `gettext` shortcut)
- Generated `.po` files for the text translations
- Added support of the `lang` argument on `/deposit` and `/withdraw`
- Added the `/language` endpoint to polaris. 
    - The intention is for @vcarl to add a "set language" button in the top right of the interactive flow pages. When changed, the client would make a request to `/language`, which would set the language for that user's session.


polaris/polaris/deposit/views.py
```diff
@@ -191,11 +144,6 @@ def deposit(account: str, request: Request) -> Response:
     except Ed25519PublicKeyInvalidError:
         return render_error_response("invalid 'account'")
 
-    # Verify the optional request arguments.
-    verify_optional_args = _verify_optional_args(request)
-    if verify_optional_args:
-        return verify_optional_args
-
```
JakeUrban: We're not doing anything with these optional args, `memo` and `memo_type`, so I removed the check. If we want to support the optional args we should file an issue and add full support for optional arguments on all endpoints as a separate effort.

JakeUrban:
This PR needs to wait until the templates have be redesigned. Then we can add translations for all the text strings in the new templates and merge this to master.

JakeUrban:
SEP-24 actually defines `lang` parameters for the `deposit`, `withdraw`, and `info` endpoints. We'll add support for these as well.

JakeUrban:
@vitchor could you fill in the Portuguese translations? You could wait until Carl finishes the templates and do them all at once if you'd like.

JakeUrban:
@vitchor there are some more strings that need translating, could you fill those in? Also, who should we talk to for spanish translations?

JakeUrban:
@vitchor actually wait a second, there are more that I got to add.

PR #77, opened 2020-01-10
Multi-Asset Support

JakeUrban:


JakeUrban:
Not yet ready. The main hurdle is rewriting `watch_transactions` to stream transactions from multiple distribution accounts. Using an Asynchronous Generator should work.

polaris/polaris/models.py
```diff
@@ -25,10 +25,7 @@ class Asset(TimeStampedModel):
     code = models.TextField(validators=[MinLengthValidator(1)], default="USD")
     """The asset code as defined on the Stellar network."""
 
-    issuer = models.TextField(
-        validators=[MinLengthValidator(56)],
-        default=settings.STELLAR_ISSUER_ACCOUNT_ADDRESS,
-    )
+    issuer = models.TextField(validators=[MinLengthValidator(56)])
```
JakeUrban: NOTE: Add migration file to remove default. 

This is necessary regardless of multi-asset support because it causes warnings about creating a migration for a new default whenever the issuer account changes.

PR #80, opened 2020-01-13
Multi-Asset Support

JakeUrban:
Adds support for anchoring multiple assets with one Polaris deployment.

Major Changes:
- `.env` should contain a list of asset codes to anchor. 
    - Each of these assets codes should also have a pair of variables for its distribution private key and issuer public key
- `watch_transactions` now uses `asyncio` to stream transactions from multiple distribution accounts instead of one
- Added a migration to remove the default for `Asset.issuer`
- Found bug in `poll_pending_deposits` which passed a stale database object to an integration function. The fix was to explicitly refresh the database object before passing it.

Small Changes:
- Removed `example` tests from circleci. 
    - This was necessary because the current setup uses the polaris source code instead of the latest published version. I'll file another issue to set up circleci to test `example` correctly. #82 
- `example` also used some settings internal to polaris, like `STELLAR_NETWORK_PASSPHRASE`. 
    - Apps using Polaris should get this value from their `.env` file.
- Adjusted tests and documentation


example/server/integrations.py
```diff
@@ -115,10 +114,9 @@ def get_stellar_toml():
         },
         "CURRENCIES": [
             {
-                "code": asset.code,
-                "issuer": polaris.settings.STELLAR_ISSUER_ACCOUNT_ADDRESS,
+                "code": "USD",
+                "issuer": "GBHJFK4LOLMYV34P25WBKDPVX75ANB3QQJCRBYO4MGWGFTUE7DBBW2PI",
```
msfeldstein: why are we hardcoding this here?

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -18,144 +21,162 @@
...
+        transaction.save()
+
+    @staticmethod
+    def _check_payment_op(
+        operation: Operation, want_asset: str, want_amount: float
+    ) -> bool:
+        return (
+            operation.type_code() == Xdr.const.PAYMENT
+            and str(operation.destination)
+            == settings.STELLAR_DISTRIBUTION_ACCOUNT_ADDRESS
```
msfeldstein: it looks like we no longer create `STELLAR_DISTRIBUTION_ACCOUNT_ADDRESS` how does this work?

example/server/integrations.py
JakeUrban: I need to change it back so good catch. I hard coded because locally, `example` uses the `polaris` source code, which no longer used `STELLAR_ISSUER_ACCOUNT_ADDRESS`.

polaris/polaris/management/commands/watch_transactions.py
JakeUrban: Good catch, it wouldn't work.

msfeldstein: do we need a test case to cover this?

JakeUrban: Yea it should have one, I'll add it.

PR #83, opened 2020-01-13
Add generic styles in line with Vega

vcarl:


vcarl:
I'm not sure where the logic for the amount input is, but I wanted to call out that it needs a `placeholder="0"` (or whatever's appropriate) to avoid getting a red error glow. The error state shows for `:invalid:not(:placeholder-shown)`, which is the best I could come up with without a more elaborate JS validation scheme.

msfeldstein:
This is all in #85, we should close right?

vcarl:
Yep, closing

Issue #78, opened 2020-01-10
Feature Request: Allow Anchors to serve content without forms


JakeUrban:
Form Integrations allows anchors to define what information needs to be collected from the user. But I've realized that the anchor may want to render non-form content to the user during the interactive flow. 

Our particular use case is that when a user submits his/her KYC info (email), the anchor should confirm that the user actually owns that email. We'd do this by sending a confirmation email and rendering a "Please confirm your email before proceeding" page until they confirm. Once they confirm, we'd move on to filling out the rest of the forms.

We should rename `form_for_transaction` to `content_for_transaction`

PR #89, opened 2020-01-16
Serve pages with no form in the interactive flow

JakeUrban:
#78 

Major Changes:
- Renamed `form_for_transaction` to `content_for_transaction`
    - This is a better name since we're returning the form _plus_ text content to be rendered in the HTML
- Changed the return type of `content_for_transaction` to a dictionary instead of the tuple
    - This is a better design since you can always expect 1 return value, not multiple or `None`
- Support serving pages without forms during the interactive flow
    - This would be needed if the user needs to take some action before continuing with the flow, like  confirming their email

### DIFF
--- a/docs/integrations/index.rst
+++ b/docs/integrations/index.rst
@@ -35,27 +35,22 @@ Form Integrations
 
 .. _Django Forms: https://docs.djangoproject.com/en/2.2/topics/forms/#forms-in-django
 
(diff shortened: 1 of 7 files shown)


JakeUrban:
Screen Shot 2020-01-16 at 4 20 28 PM

JakeUrban:
Screen Shot 2020-01-16 at 4 21 38 PM

Issue #86, opened 2020-01-15
Reference Server: Add all-fields form page for reference


JakeUrban:
We'd like to provide a page (not shown in the deposit/withdraw flow) in the reference server that has all possible form fields. This can be used as a style/content reference for other anchors.

PR #90, opened 2020-01-17
Reference Server: Added all-fields form and endpoint

JakeUrban:
#86


JakeUrban:
Screen Shot 2020-01-16 at 4 52 19 PM

JakeUrban:
Also added `<center></center>` tags to the guidance string, the picture is before that.

Issue #81, opened 2020-01-13
Correctly set up SIGNING_KEY


JakeUrban:
SEP-1 and SEP-10 don't provide any details on `SIGNING_KEY` except that it's private key should be used to sign authentication challenge transactions. @msfeldstein believes the `SIGNING_KEY` should not be a distribution or issuer account public key. This makes sense because an anchor has one signing key but could anchor multiple assets.

So we'll create a new account on testnet, set the SIGNING_KEY and SIGNING_SEED to the public and private key, respectively.

PR #96, opened 2020-01-21
Added independent signing keypair

JakeUrban:
#81


polaris/.env.example
```diff
@@ -5,6 +5,13 @@ ASSETS="USD"
 USD_DISTRIBUTION_ACCOUNT_SEED="SCRGUCH3B2NKWJUZLQWMJ2VOC3TAXGE3WW5NSNQQQZ2RX34237RZALEK"
 USD_ISSUER_ACCOUNT_ADDRESS="GBHJFK4LOLMYV34P25WBKDPVX75ANB3QQJCRBYO4MGWGFTUE7DBBW2PI"
 
+# The keypair to use to sign and verify SEP-10 authentication transactions.
+# If not specified, Polaris will use the first listed asset's distribution
+# account keypair, although you should really use a keypair independent of
+# your anchored assets.
+SIGNING_KEY="GBHS6NPDH5MZ4L4737PFBP6JQ5GHEVILLGPVXEUH5AMMNIB5IWBW6FV4"
```
msfeldstein: Since the key is derived from the seed can we only require the seed?  It would cause some confusion and issues if the key and seed don't match.  We should at least assert that up front

JakeUrban: Yea I think that makes sense.

PR #105, opened 2020-01-23
add testing harness attributes to forms

msfeldstein:
In order to have our automated validation suite be able to get through the interactive forms, we need to provide valid test-values for each of our fields.  I've added `test-value=value` and `test-action=submit` attributes to all form fields and submit buttons, which means our test suite can load a page, then find all fields that have a test-value, and set the value to that value, and then submit the form successfully.  

We should probably only show this in staging, is there a way to show or hide this attribute based on production/staging environment?


msfeldstein:
I'll do the withdraw piece when i work on the withdraw test flow

PR #106, opened 2020-01-24
gitignore sqlite db

msfeldstein:
Since we create a sqlite database when we run locally we should ignore it so it doesn't get added to the repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,5 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+polaris/db.sqlite3


JakeUrban:
@msfeldstein Actually I just realized why I had `db.sqlite3` checked in git.

Since `db.sqlite3` is a mounted volume in the container, running `RUN touch /code/db.sqlite3` won't work because it will be overwritten by `docker-compose.yml` when it tries to create the mount (it will created a `db.sqlite3` _directory_ in `polaris/`). 

So I thought the fix for that would be to run `RUN touch ./polaris/db.sqlite3` instead, so `docker-compose` finds the file and mounts it. But, it turns out that you can't create files and directories on the host from a Dockerfile. So the `db.sqlite3` needs to exist before running `docker-compose build`.

We could just adjust the documentation and tell developers to create the file before building the image. But I don't think we can create `db.sqlite3` at build time.

msfeldstein:
Yeah i was having trouble with this, i'll just file a bug for now and we can figure it out, its not pressing

Issue #60, opened 2019-12-16
Support SEP10 Multisig


tomerweller:
https://github.com/stellar/stellar-protocol/pull/489

leighmcculloch:
I'm working on a code implementation of what the protocol change will look like in our Go SDK. Recommend holding off on this until that is finished, to make sure we're all on the same page on what the change actually entails.

overcat:
When stellar/stellar-protocol#489 is merged, I will implement it in Python SDK.

JakeUrban:
@leighmcculloch is there an issue or open PR for the Go SDK change?

leighmcculloch:
No not really. The SEP-10 change is still being iterated on. It likely won't be merged without a working Go SDK change to prove it out, and both will maybe merge together. There is also a PR open for the Go SDK change, but it is very likely to be re-iterated on and changed: stellar/go#2071.

JakeUrban:
@leighmcculloch Regarding the following additional checks the server performs on the challenge transaction:
```
if the operation's source account exists:
- verify that client signatures count is one or more;
- verify that client signatures on the transaction are signers of the operation's source account;
- verify that client signatures are correct;
- verify that client signatures provide weight that meets the required threshold(s), if any;
if the operation's source account does not exist:
- verify that client signature count is one;
- verify that client signature is correct for the master key of the account address;
```

These additional requirements should really be implemented in the python SDK, specifically [verify_challenge_transaction()](https://stellar-sdk-zh-cn.readthedocs.io/en/latest/api.html#sep-0010-stellar-web-authentication). I'll make a pull request on the repository, but @overcat may not want to merge it until the SEP-10 PR is merged.

To test the SEP-10 changes then, we'd have to do one of the following:
- Test Polaris locally, running the forked python SDK
- Publish the forked python SDK publicly so we can actually deploy the changes
    - We could also deploy our own PyPI server instead, so we don't have to publish forks publicly
- Duplicate the changes made to the python SDK directly in Polaris
    - This would test the changes proposed to SEP-10 but it wouldn't reflect how it would actually be implemented in production

leighmcculloch:
Oh, I missed we had upgraded to v2 of the Python SDK and that it now had the verification being done there. I thought it was still being done 100% in Polaris.

I agree we shouldn't expect the Python SDK to accept a PR for a change that hasn't been merged in the stellar-protocol repository.

[...]

PR #107, opened 2020-01-24
SEP-10 Multisig Support

JakeUrban:
#60 

Refactors the SEP-10 authentication code and adds support for multi-signature accounts according to the changes in [stellar-protocol](https://github.com/stellar/stellar-protocol/pull/489), utilizing the `stellar_sdk` changes released in 2.1.1-beta1.

2.1.1 has been released today, so I'll update this PR to use that.


JakeUrban:
@overcat I'd like your thoughts on how this should be implemented in the Python SDK. I code comments describing my ideas.

polaris/polaris/sep10auth/views.py
```diff
@@ -89,20 +90,114 @@ def _transaction_from_post_request(request):
...
+    ##################################################################
+    # The functionality implemented below should go in stellar_sdk and
+    # invoked like so:
+    #
+    #       verify_challenge_transaction(
+    #           challenge_transaction=envelope_xdr,
+    #           server_account_id=settings.SIGNING_KEY,
+    #           network_passphrase=settings.STELLAR_NETWORK_PASSPHRASE,
+    #           account_signers=account_signers,
+    #           threshold=threshold
```
leighmcculloch: 👏 I love this design of optionally providing the `account_signers` and the `threshold` on the existing function instead of making a new function. I'm guessing for the Python SDK we can make the behavior backwards compatible by allowing the `verify_challenge_transaction` function to operate without those fields and performing the existing behavior in that case. A client can call the function without the new parameters for accounts that do not exist yet just to validate the master key signed the transaction.

polaris/polaris/sep10auth/views.py
```diff
@@ -89,20 +90,114 @@ def _transaction_from_post_request(request):
...
+    # instead of requiring the client to do some manipulation of the
+    # response before passing.
+    ##################################################################
+
+    # Get the account signers used for this transaction
+    server_kp = Keypair.from_public_key(settings.SIGNING_KEY)
+    matched_signers = []
+    for dec_signer in transaction_envelope.signatures:
+        try:
+            server_kp.verify(tx_hash, dec_signer.signature)
```
leighmcculloch: 💡 A few lines down in `_match_signer` you use the hint to guard against unnecessarily verifying signatures with the key if they won't match. You can do the same thing here.

polaris/polaris/sep10auth/views.py
```diff
@@ -89,20 +90,114 @@ def _transaction_from_post_request(request):
...
+        except BadSignatureError:
+            # dec_signer is not from the server's keypair
+            matched_signers.append(_match_signer(tx_hash, dec_signer, account_signers))
```
leighmcculloch: ❓ Will `_match_signer` append `None` to the list of `matched_signers` for any key that didn't match the server key, but did match one of the `account_signers`? Will that mean the resulting list of `matched_signers` will look something like this if there were a total of five signatures, three signers successfully matched and two were unrecognized:
```python
[
  {'keypair':'GD2WZVWK2K56Q4RWP5M2GTBHUIG56R7LWRUQMMZMK7I6TNLV72QDJEZ6', ...}, 
  {'keypair':'GD3AH24764GIOS2KUWHTIFOPRK75MB4ESLYU35JGXWKLEGKQWWEJCOJQ', ...},
  None,
  None,
  {'keypair':'GC33DJOHVMU6IVGZVDNWIMHUDP2LGARGSPBIBAV5FSESXZTPMHFHNS7V', ...}
]
```

polaris/polaris/sep10auth/views.py
```diff
@@ -89,20 +90,114 @@ def _transaction_from_post_request(request):
...
+
+    # Check threshold
+    if sum(signer["weight"] for signer in matched_signers) < threshold:
```
leighmcculloch: ❗️ It would be a good idea to deduplicate the signers, or to ensure in some way that a user signing a challenge transaction with one valid key three times can't rack up a total weight that meets a threshold. In the Go SDK I addressed that by storing the list of matched signers into a map with the key as the map key. There are a few ways you could address this in different places.

polaris/polaris/sep10auth/views.py
```diff
@@ -89,20 +90,114 @@ def _transaction_from_post_request(request):
...
+    return matched_signer
+
+
+def _get_signers_and_threshold(source_account):
+    """
+    Makes a Horizon API call to /accounts and returns the signers and
+    threshold. Replaces 'key' with 'keypair'.
+    """
+    server = settings.HORIZON_SERVER
+    account_json = server.accounts().account_id(source_account).call()
```
leighmcculloch: ❗️ We also need to handle the case that an account does not exist. I don't see that being handled here. If an account doesn't exist we can run with the master key since an account when first created has the master key as its signer and that master key must sign a transaction to make anyone else a signer. https://github.com/stellar/stellar-protocol/pull/489/files#diff-72480d0894ae2d1a31c229b64ad4c37bR135-R137

polaris/polaris/sep10auth/views.py
```diff
@@ -89,20 +90,114 @@ def _transaction_from_post_request(request):
...
-    return ""
+    except InvalidSep10ChallengeError as e:
+        raise ValueError(str(e))
+
+    transaction_envelope = TransactionEnvelope.from_xdr(
+        envelope_xdr, settings.STELLAR_NETWORK_PASSPHRASE
+    )
+    # verify_challenge_transaction ensures transaction has at least
+    # one operation whose source is the client account.
+    source_account = transaction_envelope.transaction.operations[0].source
```
leighmcculloch: 💡 It'd be nice to take this line out of users of the SDK. Users need to know to grab the source account out of the manage data operation. In the Go SDK change we're adding a separate `read_challenge_transaction` function that does much of the same verification as the `verify_challenge_transaction` but doesn't verify client signatures. It can return the source account. It also means the server doesn't need to decode the transaction at all. It might be useful to do something similar in the Python SDK.
https://github.com/stellar/go/pull/2071/files#diff-f6a8a4abae844d51cc40cb628ba55d71R417-R427

polaris/polaris/sep10auth/views.py
```diff
@@ -89,20 +90,114 @@ def _transaction_from_post_request(request):
...
     Validate the provided TransactionEnvelope XDR (base64 string). Return the
     appropriate error if it fails, else the empty string.
+
+    verify_challenge_transaction() does not verify that the transaction's
+    signers are valid and have the total weight greater or equal to the
+    source account's medium threshold value, so this function does that as
+    well.
     """
     try:
         verify_challenge_transaction(
```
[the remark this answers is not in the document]
msfeldstein: Yeah we need to add a couple test cases where 
```
Account A has signing weight
A: 0
B: 1
C: 1
and Threshold: 2
```
Case 1: Requests a token by signing with B and C succeeds, 
Case 2: Requests a token by signing with A fails
Case 3: Requests a token by signing with B fails
Case 4: Requests a token by signing twice with B fails (good catch on that leigh!)

Any other edge cases we should test for?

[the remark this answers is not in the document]
leighmcculloch: If you'd like to identify more test cases, these are all the test cases I've added to the Go SDK. You can probably flesh these out in the actual Python SDK change though, because there's a lot.

https://github.com/stellar/go/pull/2071/files#diff-74afdec610280c6cd524e507134888daR1168

This also be a good time to work on updating the [transfer server validator](https://github.com/stellar/transfer-server-validator) so that its SEP-10 test suite tests all these cases. What do you think?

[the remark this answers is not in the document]
JakeUrban: I see, so the `verify_challenge_transaction` code needs to remove the call:
```
if not __verify_te_signed_by(transaction_envelope, manage_data_op.source)

File: polaris/polaris/sep10auth/views.py (hunk 1 above)
Comment (JakeUrban) [reply polaris/polaris/sep10auth/views.py]: Exactly.

File: polaris/polaris/sep10auth/views.py (hunk 3 above)
Comment (JakeUrban) [reply polaris/polaris/sep10auth/views.py]: `_match_signer` would never return `None`. It checks if `matched_signer is None` and raises an error if it is.

File: polaris/polaris/sep10auth/views.py (hunk 4 above)
Comment (JakeUrban) [reply polaris/polaris/sep10auth/views.py]: Oh good catch, I didn't realize you could sign the transaction with the same keypair to produce multiple decorated signatures.

File: polaris/polaris/sep10auth/views.py (hunk 5 above)
Comment (JakeUrban) [reply polaris/polaris/sep10auth/views.py]: Ok, initially I thought it was a bit weird to authenticate an account that doesn't actually exist, since the eventual deposit would when it tries to submit the payment transaction. But I see [SEP-24](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#stellar-account-does-not-exist) asks anchors to create the account if it doesn't exist, which we do in `create_stellar_deposit`.

I'll implement the case where it doesn't exist.

File: polaris/polaris/sep10auth/views.py (hunk 6 above)
Comment (JakeUrban) [reply polaris/polaris/sep10auth/views.py]: Ok, I'll add a comment to add a function to extract the source account from the envelope XDR in the SDK.

File: polaris/polaris/sep10auth/views.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (leighmcculloch) [reply polaris/polaris/sep10auth/views.py]: >I think `verify_challenge_transaction` should assume the client's source account has signed the transaction _unless_ `account_signers` and `threshold` are passed.

👍

File: polaris/polaris/sep10auth/views.py (hunk 5 above)
Comment (JakeUrban) [reply polaris/polaris/sep10auth/views.py]: `verify_challenge_transaction()` already checks that the source account is a signer so if the account doesn't exist, we'll just call the function without `account_signers`.

File: polaris/polaris/sep10auth/views.py (hunk 3 above)
Comment (leighmcculloch) [reply polaris/polaris/sep10auth/views.py]: Oh, hmm. So if it encounters a signature that doesn't match any signer it fails immediately? Cool, sounds great!

**Comment by overcat** [comment]
Thanks for your suggestions, I will release a pre-release version in these two days.

**Comment by overcat** [comment]
Hi @JakeUrban, I just released a pre-release version, please let me know if you have any questions.
- [Github Release](https://github.com/StellarCN/py-stellar-base/releases/tag/2.1.1-beta1)
- [PyPi Release](https://pypi.org/project/stellar-sdk/2.1.1b1/)
- [Document](https://stellar-sdk-zh-cn.readthedocs.io/en/2.1.1-beta1/api.html#sep-0010-stellar-web-authentication)
- [Code Example](https://github.com/StellarCN/py-stellar-base/blob/2.1.1-beta1/examples/transaction_challenge.py)

File: polaris/polaris/sep10auth/views.py [hunk 8]
```diff
@@ -29,150 +35,179 @@
...
+        """
+        Generate the challenge transaction for a client account.
+        This is used in `GET <auth>`, as per SEP 10.
+        Returns the XDR encoding of that transaction.
+        """
+        return build_challenge_transaction(
+            server_secret=settings.SIGNING_SEED,
+            client_account_id=client_account,
+            anchor_name=ANCHOR_NAME,
+            network_passphrase=settings.STELLAR_NETWORK_PASSPHRASE,
```
leighmcculloch: The Python SDK still has a default timeout of 300ms rather than 900ms as in v1.3.0. I think we should set it explicitly here though so we're making a conscious decision that's what we want the timeout to be. Could we add that?

polaris/polaris/sep10auth/views.py
```diff
@@ -29,150 +35,179 @@
...
+                )
+                raise ValueError(str(e))
+            else:
+                logger.info("Challenge verified using client's master key")
+                return
+
+        signers = account.load_ed25519_public_key_signers()
+        threshold = account.thresholds.med_threshold
+        try:
+            signers_found = verify_challenge_transaction_threshold(
```
leighmcculloch: Is there a plan to do something with the `signers_found` captured from the return that got missed, or we're just putting it in the variable to remind readers that they're available?

polaris/polaris/sep10auth/views.py
```diff
@@ -29,150 +35,179 @@
...
+        logger.info(
+            f"Challenge verified, generating SEP-10 token for account {source_account}"
+        )
+        hash_hex = binascii.hexlify(transaction_envelope.hash()).decode()
+        jwt_dict = {
+            "iss": request.build_absolute_uri("/auth"),
+            "sub": source_account,
+            "iat": issued_at,
+            "exp": issued_at + 24 * 60 * 60,
+            "jti": hash_hex,
```
leighmcculloch: The `jti` is no longer required in SEP-10. It was removed in v1.2.0. stellar/stellar-protocol@353329faf6ace26b4db6d58c16bed58f23c34dc8

polaris/polaris/sep10auth/views.py
```diff
@@ -29,150 +35,179 @@
...
+        issued_at = time.time()
+        transaction_envelope, source_account = read_challenge_transaction(
+            envelope_xdr, settings.SIGNING_KEY, settings.STELLAR_NETWORK_PASSPHRASE
+        )
+        logger.info(
+            f"Challenge verified, generating SEP-10 token for account {source_account}"
+        )
+        hash_hex = binascii.hexlify(transaction_envelope.hash()).decode()
+        jwt_dict = {
+            "iss": request.build_absolute_uri("/auth"),
```
leighmcculloch: I think this line allows injection into the JWT. If an anchor hosts their server in anyway that allows arbitrary hostnames in the requests, e.g. if they host it on a static IP without hostname filtering which is possible, the `iss` field can be made to contain any hostname a caller wants.

I'd lean towards this being the server signing key address, e.g. https://github.com/stellar/go/blob/4268b290621430b89c3ed687e0d83df3800c227a/exp/services/webauth/internal/serve/token.go#L87, if we don't have a good way to get some other definitive exact URI for the issuer.

polaris/polaris/sep10auth/views.py
JakeUrban: Its used in the log statement, but there is no plan to use it functionally. I'm not particularly attached to having that log statement though so if you think its not needed I can remove it.

leighmcculloch: Ah, no worries. Sorry I missed the log statement. It's fine to leave, I just wasn't sure if there was something incomplete, but it looks good to me.

polaris/polaris/sep10auth/views.py
JakeUrban: I'm not convinced that using the server signing key is any better than the current value for `iss`; both can be faked. The only real protection against someone attempting to side-step SEP-10 by using a fake JWT (a JWT not created by the anchor server) is the server's secret used to encode the data. The SEP also states that `iss` should be a URI. What do you think?

leighmcculloch: Ah you're right, the SEP does say to use a URI. I was thinking about the RFC7519 Section 4.1.1 where technically it doesn't have to be a URI, but I forgot we specify that in the SEP. I'll change the `webauth` implementation to make it a URI too.

Can we get the hostname from the settings or somewhere else instead of the incoming request? I still think we should avoid allowing any data from the request making its way into the JWT.

leighmcculloch: FYI I updated the Go implementation here to align with ☝️: stellar/go#2214.

polaris/polaris/sep10auth/views.py
leighmcculloch: I'm going to resolve this comment. Because Polaris is already in use it might not be safe to remove the `jti` and it's not really relevant to the change you're trying to make in this PR. It doesn't hurt to keep it.

polaris/polaris/sep10auth/views.py
JakeUrban: Yea, I can add an `AUTH_SERVER` setting since the SEP-10 authentication could be hosted on a different URI than the anchor server.

JakeUrban: Actually, if they're using Polaris there isn't a reason you'd want to have a separate server handling authentication. I'll use a `HOST_URL` instead.

leighmcculloch:
>I haven't finished looking through the Python SDK challenge transaction implementation that is in 2.1.1, I'll post here when I've done that too.

FYI I had a look through the Python SDK challenge transaction implementation that is in 2.1.1 and I posted some recommendations on that repositories issue about. It would be really worthwhile if the changes to how signatures and signers are matched could be made since we're using this in Polaris.
https://github.com/StellarCN/py-stellar-base/issues/252#issuecomment-580882560

Issue #110, opened 2020-01-27
Feature Request: Support `callback` argument for the interactive flows


JakeUrban:
SEP-24 defines the use of a `callback` argument when making a request to an anchor's interactive URL, but Polaris doesn't currently support it. The SEP doesn't indicate that its optional, but this issue is a reminder to follow up with the team on whether or not we should make it more clear.

PR #112, opened 2020-01-28
Support 'callback' parameter for interactive flows

JakeUrban:
#110


JakeUrban:
Still need to adjust docs, add a couple tests

polaris/polaris/helpers.py
```diff
@@ -259,15 +260,17 @@ def invalidate_session(request: Request):
 
 def interactive_args_validation(
     request: Request,
-) -> Tuple[Optional[Transaction], Optional[Asset], Optional[Response]]:
+) -> Tuple[Optional[Transaction], Optional[Asset], Optional[str], Optional[Response]]:
```
msfeldstein: Can we make this a type instead of a tuple so that we don't need to change all the caling sites everytime something is added or removed, and it will also be self-documenting since we'd have a `.callback` property instead of an `Optional[str]` property

JakeUrban:
I need to step back and rethink the approach. Originally I thought we could make the callback request on the client-side from the `more_info.html` template, but that won't work. Here's why.

`more_info.html` is rendered in 2 situations:
- `POST /interactive` redirects to `more_info.html` when all the transaction information has been collected. At this point, the status will be `pending_user_transfer_start`.
- When its requested directly, i.e. `GET /transaction/more_info`

Because the demo client repeatedly hits `GET /transaction/more_info`, I was under the impression that `more_info.html` is rendered whenever the status changes, but that won't be the case for most wallets.

_There is no guaranteed that `more_info.html` will be rendered at when the transaction has been completed._

Therefore, we cannot make `callback` requests from the client-side. These `callback` requests _have to_ be made on the server side, once `poll_pending_deposits` or `watch_transactions` determines a submitted transaction has completed successfully.

JakeUrban:
Considering the above points, `callback` URL's need to persist, most likely in the database, on a per-transaction basis. It cannot be stored in a web session, because the `poll_pending_deposits` and `watch_transactions` processes won't have access to the session key.

This likely merits a new column on the `Transaction` model, or a one-to-one table, like `TransactionCallback`.

polaris/polaris/templates/transaction/more_info.html
```diff
@@ -88,22 +105,22 @@
...
-            //         "id": "fake-transaction-id",
-            //         "status": "completed"
-            //         etc.
-            //     }
-            // }
             targetWindow.postMessage(tx_json, "*");
         }
+
+        // Callback function to post the serialized transaction to the callback URL
+        function urlCallback() {
```
msfeldstein: i think this should be done server side, like we talked about this morning then we can have immediate callbacks regardless of whether this page is open in the client or not.

msfeldstein: looks like this actually matches the sep in that it only communicates that the interactive flow is finished, so np here

PR #113, opened 2020-01-30
Fee Integrations

JakeUrban:
#117 

Changes:
- Added a `fee_func` argument to `register_integrations`
- Defined the default `fee_func` as `calculate_fee` (similar to `calc_fee`)
- Adjusted `Asset` model default values, added a migration file
- Removed `fee_*` attributes from `/info` response
- Removed check for unsupported `op_type`

`registered_fee_func` is used in 2 different cases:
- deposit/withdrawal interactive flow, after a `TransactionForm` has been submitted
- Requests to the `/fee` endpoint

Parameters passed to `registered_fee_func` for `/fee` requests are the same parameters the `/fee` endpoint accepts: `asset_code`, `operation`, `type`, `amount`.

After `TransactionForm` submissions, `asset_code`, `operation`, and `amount` are still passed to `registered_fee_func`. The difference is that every other field defined for the `TransactionForm` subclass will also be passed like so:
```
    if form.is_valid():
        if is_transaction_form:
            # Pass `operation`, `asset_code`, and `amount` to registered fee
            # function, as well as any other fields on the TransactionForm.
            # Ex. operation `type`
            fee_params = {
                "operation": settings.OPERATION_WITHDRAWAL,
                "asset_code": asset.code,
                **form.cleaned_data,
            }
            transaction.amount_in = form.cleaned_data["amount"]
            transaction.amount_fee = registered_fee_func(fee_params)
            transaction.save()
```
Anchors can define a `type` field on their `TransactionForm` subclass if fee calculation requires it. If anchors don't want to actually display specific fields but still want them passed to `registered_fee_func`, they can use the [hidden form fields](https://docs.djangoproject.com/en/3.0/ref/forms/widgets/#hiddeninput).


polaris/polaris/migrations/0006_auto_20200129_2323.py
```diff
@@ -0,0 +1,114 @@
...
+                ],
+            ),
+        ),
+        migrations.AlterField(
+            model_name="asset",
+            name="deposit_max_amount",
+            field=models.DecimalField(
+                blank=True,
+                decimal_places=7,
+                default=10000000000000000000000,
```
msfeldstein: https://docs.python.org/3/library/decimal.html#decimal.MAX_EMAX

Decimal.MAX_EMAX ?

polaris/polaris/models.py
```diff
@@ -22,7 +25,7 @@ class Asset(TimeStampedModel):
     This defines an Asset, as described in the SEP-24 Info_ endpoint.
     """
 
-    code = models.TextField(validators=[MinLengthValidator(1)], default="USD")
```
msfeldstein: Why remove this?

docs/index.rst
```diff
@@ -162,7 +162,7 @@ intend to anchor. Get into the django python shell like so:
 ::
 
     from polaris.models import Asset
-    Asset.objects.create(code="USD", issuer="<the issuer address>")
+    Asset.objects.create(...)
```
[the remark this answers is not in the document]
JakeUrban: No it doesn't but I can add them.

polaris/polaris/migrations/0006_auto_20200129_2323.py
JakeUrban: Nice catch, I'll change that.

polaris/polaris/models.py
JakeUrban: The column is already `NOT NULL` (`null=False` by default in django), so I felt that the `MinLengthValidator` was hand holding. The anchor would have to explicitly create the asset with an empty string as the `code` since `USD` is the default.

PR #126, opened 2020-02-06
check_trustlines() now checks asset_issuer and limit as well

yuriescl:
There can be multiple assets with the same asset code from different issuers, so the issuer needs to be checked as well.
The account might contain balance for an asset but not have the trustline, in this case the balance is 0 and limit is 0. A limit higher than 0 means the trustline exists.

### DIFF
--- a/polaris/polaris/management/commands/check_trustlines.py
+++ b/polaris/polaris/management/commands/check_trustlines.py
@@ -73,6 +73,7 @@ class Command(BaseCommand):
                     asset_issuer = balance["asset_issuer"]
                     limit = float(balance["limit"])
                 except KeyError:
+                    logger.debug(f'balance {balance}')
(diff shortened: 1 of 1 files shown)


yuriescl:
I just changed that section of the code but I'm not familiar with your testing code, so I didn't updated it and that's why it's failing. Do you want me to update it as well? (might take a few days)

yuriescl:
I think my code has some issues, working on it.

yuriescl:
That solution works for an older version of stellar sdk, I'm closing this Pull Request and creating and issue instead since I'm not familiar with the new sdk.

yuriescl:
Nevermind this, I think the new sdk doesn't list the asset in balances if there's no trustline. Since I'm using v1.4 of stellar sdk, I had to add the issuer/limit check but I don't think this is needed here.

PR #129, opened 2020-02-06
Check issuer for in check_trustlines CLI

JakeUrban:
@yuriescl any comments and suggestions are welcome.


yuriescl:
@JakeUrban Oh I didn't see you were generating the balances response manually, that's why I thought the sdk was returning only the asset code. Looks good to me, thanks for the fix!

Issue #79, opened 2020-01-11
Feature Request: Add confirm email step for new users


JakeUrban:
As described in #78 , we want to allow anchors to serve pages during the interactive flow without forms. We'll use this feature to add a "confirm email" step for new users. 

Polaris users are identified by an email. If the stellar account passed from the wallet is already associated with an email, KYC has already been collected. If the stellar account is new to the anchor, we'll render the KYC form for the user to fill out.

PR #132, opened 2020-02-07
Confirm email step

JakeUrban:
resolves stellar/django-polaris#79 

Flow:
1. User submits KYC info (name, email)
Screen Shot 2020-02-07 at 9 43 40 AM

2. The response:
Screen Shot 2020-02-07 at 9 43 50 AM

3. Check email (this may be in spam folder.. not sure how to fix that):
Screen Shot 2020-02-07 at 9 44 07 AM

4. Select "Confirm Email", which takes you to a confirmation page:
Screen Shot 2020-02-07 at 9 44 25 AM

5. Bring the popup back into focus, it should render the next form:
Screen Shot 2020-02-07 at 9 44 35 AM


JakeUrban:
The reason I chose to use the title is because its the only per-page attribute the integration points allow you to control. Polaris allows you to change the page title based on the return value from `content_for_transaction()`. I can't think of another way someone using Polaris would be able to  take some action on the client side for a particular page or state. Open to suggestions.

PR #137, opened 2020-02-12
Trying to trigger workflow

JakeUrban:


JakeUrban:
It appears that workflows cannot be triggered from forked repositories:
https://github.community/t5/GitHub-Actions/Github-Workflow-not-running-from-pull-request-from-forked/td-p/32979

This is a disappointment to say the least, especially considering that most developers and organizations create PR's from forked repositories. I'm going to try and make a PR from stellar/django-polaris to make sure it works there.

PR #138, opened 2020-02-12
Try to Trigger a Workflow

JakeUrban:
I created a workflow file but it wasn't running when I made a PR from my forked repository. I found out that workflows are pretty new and don't currently support events originating from forks. So, I'm creating this PR to determine if the workflow is triggered when the PR originates from the same repository.

If this is successful, we may want to change our git flow so we can use it.

### DIFF
--- a/polaris/polaris/stellartoml/views.py
+++ b/polaris/polaris/stellartoml/views.py
@@ -12,6 +12,8 @@ from django.http.response import HttpResponse
 from polaris import settings
 from polaris.integrations import registered_toml_func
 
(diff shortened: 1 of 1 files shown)


JakeUrban:
Here is the issue related to workflows not running from forked repo's: https://github.community/t5/GitHub-Actions/Github-Workflow-not-running-from-pull-request-from-forked/td-p/32979

However my workflow didn't run after creating this PR, so I'm not sure what the issue is at the moment.

Issue #134, opened 2020-02-10
Allow per-page scripts integration


msfeldstein:
We have a script we use in our email confirmation form page to reload the page when you focus it (return to the app after confirming email).  Currently we're using [the title of the html page](https://github.com/stellar/django-polaris/pull/132/files#diff-b3845ebdab78f3325c727162cf116d73R334) to determine when to run the script, but a cleaner way would be to only inject that script to the page where its needed.

One possible solution would be to have the `scripts()` integration take a page key, and have the Dict returned by `content_for_transaction` return a page_key that `scripts(page_key)` can use to determine which scripts to attach

JakeUrban:
I like the idea but I'm hesitant to add another explicitly defined key to `content_for_transaction()`. What do you think about passing the entire dict to `scripts()` instead a particular key value? That way, the scripts integration function could tailor the scripts returned based on the form or any extra keys the developer decides to return.

msfeldstein:
i like that idea.

PR #140, opened 2020-02-12
Pass page_content to registered script func

JakeUrban:
resolves stellar/django-polaris#134 

Also changes `registered_javascript_func` to `registered_scripts_func` and the `register_integrations()` `javascript_func` parameter to `scripts_func`.

### DIFF
--- a/example/server/integrations.py
+++ b/example/server/integrations.py
@@ -313,8 +313,8 @@ def get_stellar_toml():
     }
 
 
-def scripts():
-    return [
+def scripts(page_content: Optional[Dict]):
+    tags = [
         # Google Analytics
         """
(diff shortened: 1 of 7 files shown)


example/server/integrations.py
```diff
@@ -325,23 +325,24 @@ def scripts():
           gtag('js', new Date());
           gtag('config', 'UA-53373928-6');
         </script>
-        """,
+        """
+    ]
+    if "form" not in page_content and page_content.get("title") == _("Confirm Email"):
```
msfeldstein: This seems very fragile to have the string `Confirm Email` duplicated, if we change it above this will brack.  Can we use a constant or something?

PR #141, opened 2020-02-13
More Info Authentication

JakeUrban:
resolves stellar/django-polaris#142

Adds authentication to /more_info requests

Other changes:
- No longer invalidates sessions once they've completed the interactive flow. This was necessary since they still need to be authenticated to access /more_info
- Fixes a bug for /interactive/complete: It requires authentication, which in turn requires a `transaction_id` argument in the URL, not `id`.
- Removes `check_middleware()` usage in tests and import statements. This check was moved out of the request cycle and added to the startup process earlier.


JakeUrban:
We decided to leave more_info unprotected at the moment. This may be revisited later.

Issue #150, opened 2020-02-18
Feature Request: form initialization parameters in integrations.py


ronaldoy:
### What problem does your feature solve?
Some forms require parameters for the _init_ section. Currently, there is no way to pass them using the return function in integrations.py. 

### What would you like to see?
Possibility to pass arguments in the "form" section of the "return" function (integrations.py) so the form can be initialized. For example, in Django it is possible to reference a form as SampleForm(param1=var1) and use var1 during the init. Any structure where we can replicate this behavior would solve the problem.

JakeUrban:
Originally, I wanted to find a solution that would allow anchor's to initialize the form themselves inside `content_for_transaction()`, but this would require anchor's to perform logic currently handled by Polaris: initializing the form with URL or POST body arguments passed in the request.

This includes the optional `amount` argument used to pre-populate a form and also the data submitted via a previously rendered form. `amount` should also only be used to initialize the form when it is a `TransactionForm`, further complicating the logic every anchor would have to implement.

Instead, I believe the best approach is to return the form and it's initialization parameters as a tuple. Polaris would still handle the logic described above _and_ initialize the form with any key-value pairs passed in the arguments dictionary:
```
def content_for_transaction(transaction):
    return {
        "form": (MyForm, {"arg1": "val1"}),
    }
```
Polaris would then initialize the form with the argument dictionary like so:
```
content = rdi.content_for_transaction(transaction)
if content.get("form"):
    form_class, args = content["form"]
    content["form"] = form_class(**args)
```
This would get the job done but it makes the return type of `content_for_transaction()` more complex.

PR #158, opened 2020-02-24
Allow anchor to pass form parameters for initialization

JakeUrban:
resolves stellar/django-polaris#150 

nTokens needs to be able to pass their own parameters to the forms served to users during initialization, and this was not previously possible since Polaris initialized the form with URL or POST arguments.

With this change, `content_for_transaction()`'s `form` key value changes from a `forms.Form` class to a tuple containing the form class as well as a dict containing any keyword arguments the anchor wants to use during initialization.

@ronaldoy for reference

### DIFF
--- a/polaris/polaris/deposit/views.py
+++ b/polaris/polaris/deposit/views.py
@@ -88,8 +88,17 @@ def post_interactive_deposit(request: Request) -> Response:
             content_type="text/html",
         )
 
-    form_class = content.get("form")
-    form = form_class(request.POST)
+    try:
+        form_class, form_args = content.get("form")
+    except TypeError:
+        logger.exception("content_for_transaction(): 'form' key value must be a tuple")
+        return render_error_response(
+            _("The anchor did not provide content, unable to serve page."),
+            status_code=500,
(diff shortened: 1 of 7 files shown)


polaris/polaris/withdraw/views.py
```diff
@@ -125,7 +134,7 @@ def post_interactive_withdraw(request: Request) -> Response:
 
     else:
         content.update(form=form)
-        return Response(content, template_name="withdraw/form.html")
+        return Response(content, template_name="withdraw/form.html", status=400)
```
msfeldstein: 422 here as well

polaris/polaris/deposit/views.py
```diff
@@ -128,7 +137,7 @@ def post_interactive_deposit(request: Request) -> Response:
 
     else:
         content.update(form=form)
-        return Response(content, template_name="deposit/form.html")
+        return Response(content, template_name="deposit/form.html", status=400)
```
msfeldstein: This should be 422, 400 is for broken inputs, 422 is for invalid input.

PR #166, opened 2020-03-05
Fix typo in PT translations

vitchor:


JakeUrban:
This file has to be re-compiled in order for these changes to take effect.

PR #171, opened 2020-03-25
return 422 on extra and unprocessable requests

### LINKED ISSUE #170: Return 4XX status on extra GET/POST requests for interactive forms
It looks like Vega may be making extra POST requests to the `/interactive` endpoints when the anchor (nTokens) doesn't have another form to serve. The response is a 500 error, but its not really the fault of the anchor, instead the client shouldn't be making those requests.

I'll find the appropriate 4XX status Polaris should return in this case.

JakeUrban:
resolves #170 

*not ready to be merged, but ready for review*

Returns 422 when the anchor doesn't return content for the interactive flow response AND the transaction is not in `incomplete` status.

@vitchor Can you send me the translation for the error message added?


vitchor:
LGTM

PR #173, opened 2020-04-01
Fix typo in Portuguese django.po translation doc

vitchor:


JakeUrban:
We should probably remove the `R$` right?

PR #174, opened 2020-04-02
Add numeric keypad for amount field

ronaldoy:
Currently the keypad in iOS shows small numbers with some special characters. This commit changes to a keypad with large numbers, improving user experience on mobile devices.


vitchor:
LGTM

PR #175, opened 2020-04-03
Minimum/Maximum amount error messages

ronaldoy:


JakeUrban:
@ronaldoy I'll likely be releasing a new version in the next week or two.

PR #179, opened 2020-04-13
SEP-9 Support

### LINKED ISSUE #164: Feature Request: accept SEP-9 params to pre-fill KYC form
If the wallet sends SEP-9 params in the deposit/withdraw endpoints, we should pre-populate the KYC fields (so users don't have to type that twice).

Here are some SEP-9 fields, as an example: email_address, first_name, last_name and mobile_number.

JakeUrban:
resolves #164 

Added an integration point for anchors to save the SEP-9 fields passed to the `/interactive` endpoints by wallets. 

The anchors are expected to save this information in a model they've defined. Then, for each form requested during the interactive flow, the anchor should return the necessary fields related to the form also returned.

### DIFF
--- a/docs/integrations/index.rst
+++ b/docs/integrations/index.rst
@@ -41,7 +41,7 @@ The functions below facilitate the process of collecting the information needed.
 
 .. autofunction:: polaris.integrations.WithdrawalIntegration.after_form_validation
 
-Some wallets may pass fields documented in SEP-9_ to the `/interactive` endpoints for
+Some wallets may pass fields documented in SEP-9_ in the initial POST request for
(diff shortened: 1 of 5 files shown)


docs/integrations/index.rst
```diff
@@ -40,6 +41,14 @@ The functions below facilitate the process of collecting the information needed.
 
 .. autofunction:: polaris.integrations.WithdrawalIntegration.after_form_validation
 
+Some wallets may pass fields documented in SEP-9_ to the `/interactive` endpoints for
```
msfeldstein, line 44: `to the /interactive endpoints` makes it sound like you're passing it to the webapp.  Maybe say `to the initial POST request`

PR #182, opened 2020-04-14
Changed content_for_transaction API

JakeUrban:
This change is in preparation for #179 , which will allow anchors to prefill their forms with parameters from SEP-9. The intention is to simply the flow for prefilling a form's fields.

**Problem Statement**
Without this change, (as #179 is implemented now), anchors must return a dictionary of initialization arguments from `content_for_transaction()` that will then be passed to the form's constructor by Polaris.

This is not a clean design because this dictionary would now contain two types of data: data that is specific to the anchor's logic (the original purpose of the design), and SEP-9 parameters for pre-filling the form. Anchors would need to separate the two types of data from the original dictionary and call `super().__init__(initial=prefill_data)` in the form's `__init__()` function.

**Solution**
Instead, we can give the anchor more control by allowing them to initialize their own forms, removing the need for them to return initialization arguments from `content_for_transaction()` and define an `__init__()` to call `super().__init__(initial=)`.

**Drawbacks**
The downside of this change is its a breaking change, specifically for `content_for_transaction()`
1. Changing the returned dict's "form" key from a tuple to an initialized form object
2. Adding two keyword arguments to `content_for_transaction()`
    - `post_data`: the data POSTed by the wallet that the anchor will now need to pass to the form's constructor when non-`None`
    - `amount`: the amount the wallet sent in the GET request for the form. The anchor should use it to prefill its `TransactionForm` if non-`None`

The second drawback adds complexity to the anchor's `content_for_transaction()` implementation, but overall I think this simplifies the API and the data flow.

### DIFF
--- a/docs/integrations/index.rst
+++ b/docs/integrations/index.rst
@@ -25,7 +25,7 @@ validate, and process the information you need to collect from users with
 `Django Forms`_.
 
 Of course, you'll need to collect the amount the user would like to deposit
-or withdraw. Polaris provides a :class:`TransactionForm` that can be subclassed to
+or withdraw. Polaris provides a ``TransactionForm`` that can be subclassed to
 add additional fields for this purpose. One ``TransactionForm`` should be rendered
 for every transaction processed. See the :doc:`../forms/index` documentation for
 more information.
--- a/example/server/integrations.py
+++ b/example/server/integrations.py
@@ -99,7 +99,7 @@ def track_user_activity(form: forms.Form, transaction: Transaction):
     )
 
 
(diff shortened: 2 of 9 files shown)


JakeUrban:
The docs and reference server need to be updated still, but I'm going to wait until we decide if we want to go with this design.

JakeUrban:
@ronaldoy What do you think of the design? Is it clear when you would have to pass the `post_data` to the form's constructor or when you would need to pass the `amount` dictionary to initial?

ronaldoy:
@jakeurban Yes, it is clear.

PR #183, opened 2020-04-15
Fixed TransactionForm bug, add min and max to label

### LINKED ISSUE #169: Add min/max amount for asset in deposit/withdraw
We validate the amount specified for a deposit or withdrawal is within the limits but it would be nice if the user saw the limits before attempting to submit the amount.

JakeUrban:
resolves #169 

I originally wanted to add the limits to the placeholder, but that attribute is not allowed on decimal fields per HTML5.

When I was adding the limits I realized that the TransactionForm class didn't know what kind of transaction it was (deposit or withdrawal) and therefore couldn't know the correct minimum and maximum. Previous versions of Polaris were validating withdraw transactions against deposit limits.

I added the limits to the `Amount` label, so it looks like this:
Screen Shot 2020-04-15 at 2 59 20 PM

However, if there are upper and lower limits, its possible the label wraps:
Screen Shot 2020-04-15 at 3 00 38 PM

One solution is to use the `<` and `>` symbols instead of `minimum` and `maximum`

This also introduces breaking change to `TransactionForm`, requiring a `Transaction` object be passed instead of an `Asset` class.


msfeldstein:
Do we have a currency Symbol available?  Can we do
"Amount (Up to $1000)"
"Amount ($10 - $10000)"
"Amount (Minimum $10)"

JakeUrban:
@msfeldstein I don't think we can use symbols because the symbol depends on the asset being deposited, so the anchor would have to give us the symbol somehow.

But I like the `-` usage.

Issue #178, opened 2020-04-13
SEP-6 Support


JakeUrban:
Since we've brought back SEP-6, we should support it in the same way we support SEP-24. This issue is to explore how to best add that support.

JakeUrban:
There were two options I considered. The first was simply making the endpoints available on the Polaris server based on an `ACTIVE_SEPS` setting, and the second was to create separate Django apps for each SEP.

While the second was a "cleaner" design, it would've made it more difficult to reuse shared functionality and taken more time and effort to refactor the code. So, I chose to go with the first option.

SEP-6 and SEP-24 endpoints will be namespaced under `/sep6` and `/sep24` URL paths. This will help make it clear what protocol the client is using and avoids functionality conflicts for the shared endpoints such as `/info` and `/transactions`.

PR #187, opened 2020-04-21
SEP-6 Support

JakeUrban:
resolves stellar/django-polaris#178

[New Docs](https://html.jakeurban.now.sh/)

Refactors Polaris to allow developers to deploy each supported SEP independently. This means users can deploy any combination of SEP 1, 10, 6, 12, and 24. For a full overview of the functionality and integrations check out the docs linked above.


polaris/polaris/sep12/customer.py
```diff
@@ -0,0 +1,46 @@
...
+        return render_error_response(str(e), status_code=400)
+    else:
+        return Response({}, status=202)
+
+
+@api_view(["DELETE"])
+@renderer_classes([JSONRenderer])
+@validate_sep10_token("sep6")
+def delete_customer(account_from_auth: str, request: Request, account: str) -> Response:
+    not_found = render_error_response("account not found", status_code=404)
```
msfeldstein, line 38: feels odd to define a not_found error up here before we need it, especially since its called `render` which makes it sound liek it actually has side effects.  Can we just return it inline below?  The error message for line 39 might also be different since its not technically a not found error

polaris/polaris/sep24/serializers.py
```diff
@@ -20,9 +23,18 @@ def get_more_info_url(self, transaction_instance):
         if not request_from_context:
             raise ValueError("Unable to construct url for transaction.")
 
-        path = reverse("more_info")
-        path_params = f"{path}?id={transaction_instance.id}"
-        return request_from_context.build_absolute_uri(path_params)
+        if "sep6" in self.context.get("request").build_absolute_uri():
+            path = None
+            if "sass_processor" in settings.INSTALLED_APPS:
```
msfeldstein, line 28: What does having sass installed have to do with the logic here?

polaris/polaris/sep24/serializers.py
```diff
@@ -68,6 +80,9 @@ def to_representation(self, instance):
         else:  # withdrawal
             del data["deposit_memo"]
             del data["deposit_memo_type"]
+        if not self.context.get("sep6"):
```
msfeldstein, line 83: Why would a SEP6 object ever go through a sep24 serializer?

polaris/polaris/sep6/transaction.py
```diff
@@ -0,0 +1,52 @@
...
+@api_view(["GET"])
+@renderer_classes([JSONRenderer])
+@validate_sep10_token("sep6")
+def transactions(account: str, request: Request) -> Response:
+    """
+    Definition of the /transactions endpoint, in accordance with SEP-0006.
+    See: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#transaction-history
+    """
+    if account != request.GET.get("account"):
+        print(account, request.GET.get("account"))
```
msfeldstein, line 36: remove

polaris/polaris/sep6/transaction.py
```diff
@@ -0,0 +1,52 @@
...
+        return render_error_response(
+            _("The account specified does not match authorization token"),
```
msfeldstein, line 38: Do we not need this check in sep24?

polaris/polaris/sep6/urls.py
```diff
@@ -0,0 +1,16 @@
...
+
+urlpatterns = [
+    path("deposit", deposit.deposit),
+    path("withdraw", withdraw.withdraw),
+    path("info", info.info),
+    path("fee", fee.fee),
+    path("transaction", transaction.transaction),
+    path("transactions", transaction.transactions),
+]
+if "sass_processor" in settings.INSTALLED_APPS:
```
msfeldstein, line 13: whats the meaning of sass_processor as app logic?

polaris/polaris/sep6/utils.py
```diff
@@ -0,0 +1,41 @@
+from typing import Dict, Optional
+
+from polaris.utils import Logger, SEP_9_FIELDS
+from polaris.integrations import registered_customer_integration as rci
+
+
+logger = Logger(__name__)
+
+
+def validate_403_response(account: Optional[str], integration_response: Dict) -> Dict:
```
msfeldstein, line 10: Comment what this is used for

docs/index.rst
```diff
@@ -5,75 +5,71 @@ Introduction
...
         "corsheaders",
         "rest_framework",
-        "sass_processor",
         "polaris",
     ]
 
-Add Polaris' :doc:`PolarisSameSiteMiddleware </middleware/index>`,
-and ``CorsMiddleware`` to your ``settings.MIDDLEWARE``.
-``SessionMiddleware`` must be listed `below` ``PolarisSameSiteMiddleware``.
+Add ``CorsMiddleware`` to your ``settings.MIDDLEWARE``. It should be listed above
```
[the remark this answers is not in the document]
JakeUrban: Yes, its possible using a django signal the package provides: https://github.com/adamchainz/django-cors-headers#signals

However I need to test how this works. If we add a signal to allow all origins for Polaris URLs, the user would have to explicitly list origins that are acceptable to all other endpoints. What if the user has another set of URLs that should allow all origins? 

If it is possible to register multiple signals so users could define additional filters then this works. But if only one signal function can be registered to accept or deny origins, it probably makes sense for the user to write it themselves. Otherwise we'd be requiring them to explicitly whitelist origins or find a workaround.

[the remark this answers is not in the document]
JakeUrban: Either way, the user will still need to add `CorsMiddleware` so Polaris can allow all origins.

Let me experiment with different configurations. For now, since this is already how it works, I don't think it should block this PR.

polaris/polaris/sep10/utils.py
```diff
@@ -4,31 +4,38 @@
...
     Check SEP 10 authentication in a request.
     Else call the original view function.
     """
     try:
         account = validate_jwt_request(request)
     except ValueError as e:
-        return render_error_response(
-            str(e), content_type=content_type, status_code=status.HTTP_403_FORBIDDEN
-        )
+        if sep == Transaction.PROTOCOL.sep6:
```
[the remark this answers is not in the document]
JakeUrban: SEP-24 isn't opinionated about what the response body should look like for 403 permission denied responses. But SEP-6 explicitly gives the response body that should be returned: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0006.md#5-authentication-required

polaris/polaris/models.py
```diff
@@ -46,7 +46,7 @@ class Asset(TimeStampedModel):
 
     # Deposit-related info
     deposit_enabled = models.BooleanField(default=True)
-    """``True`` if SEP-6 deposit for this asset is supported."""
+    """``True`` if deposit for this asset is supported."""
```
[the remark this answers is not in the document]
JakeUrban: Thats true. We could change `deposit_enabled` to `sep24_deposit_enabled` and `sep6_deposit_enabled`, and same thing for withdraw. Then we wouldn't need the `sep24_enabled` and `sep6_enabled` columns.

I don't see any drawbacks and it provides more flexibility so I'll do it.

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -200,18 +200,22 @@ def update_transaction(
...
         code = operation.asset.code
-        print(want_asset, want_amount)
-        print(operation.destination, want_asset.distribution_account)
-        print(str(issuer), want_asset.issuer)
+
+        # SEP-6 doesn't require 'amount' parameter when creating transaction
+        # so we can't use it to match the transaction. Only use it if present.
+        amounts_match = True
+        if want_amount:
+            amounts_match = want_amount == Decimal(operation.amount)
```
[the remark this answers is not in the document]
JakeUrban: Yea that makes sense. This gets into a whole area of tough situations. 

If we remove the check and someone withdraws _way more_ than they specified in the interactive flow, it would be up to the anchor to notice the difference in amounts and raise an error. If they didn't catch it and just sent the amount the user specified in the interactive flow, the user just lost a chunk of their money right?

I guess this is what currently happens when they send an amount that doesn't match, but Polaris should probably handle this case better.

I'll file an issue to explore the options. Maybe we provide more flexibility in the integration to allow anchors to reject withdraws and send them back?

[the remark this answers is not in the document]
JakeUrban: https://github.com/stellar/django-polaris/issues/196

polaris/polaris/sep24/serializers.py
JakeUrban: Polaris the UI pages require that package. SEP-6 doesn't require any UI pages, but `more_info_url` is optional. So if they want `more_info_url`, they have to have this installed, which they might not.

Thinking about this more I'd like to remove all the SASS stuff. We can compile all the SASS files offline before each release and reference the CSS files in the templates. This will simply the configuration too.

Filing an issue for it. For now, the docs describe when a Polaris user would need to have this installed.

polaris/polaris/sep24/serializers.py
JakeUrban: The responses are identical except for these two attributes. I could create a separate serializer but I figured its not worth it since the attributes they they won't change.

polaris/polaris/sep6/transaction.py
JakeUrban: Nope, SEP-24 doesn't accept the `account` parameter, its expected to extract it from the auth token.

polaris/polaris/sep6/urls.py
JakeUrban: Explained in the related comment. UI pages require `sass_processor` but UI pages (more_info_url) are optional in SEP-6.

docs/index.rst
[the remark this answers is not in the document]
msfeldstein: ok this isn't relevent to this PR no need to make it bigger, but something to consider for long term maintenance

polaris/polaris/sep10/utils.py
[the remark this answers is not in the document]
msfeldstein: If SEP-24 isn't opinionated why not just always use the sep6 response?

polaris/polaris/sep24/serializers.py
msfeldstein: If this is shared then we should put it in a shared folder

polaris/polaris/sep6/urls.py
msfeldstein: so we're using the existence of the sass processor as an indicator that there is a more_info_url?  Can we do something more explicit? If we stop using sass processor then this logic will break

polaris/polaris/sep24/serializers.py
msfeldstein: Just add a comment here explaining what the if statement implies, and then we have a task to remove sass altogether

polaris/polaris/sep6/urls.py
JakeUrban: The requirement of having `sass_processor` installed will be removed before the next release. When that happens we can remove these conditions.

polaris/polaris/sep24/serializers.py
JakeUrban: done

polaris/polaris/sep10/utils.py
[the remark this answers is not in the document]
JakeUrban: This isn't a strongly held opinion but I think using this response for SEP-24 doesn't really make as much sense. 

The `{'type': 'authentication_required'}` response makes sense in SEP-6 because its accompanying status code (403) is used for 3 other response types. Thats why SEP-6 has a `type` attribute in all 403 responses, to distinguish what kind of 403 response it is.

For all SEP-24 endpoints, error responses include an `error` key. If we make an exception for bad authentication responses clients may not be expecting a different schema. Again, this isn't a big deal but I think its better to be consistent.

docs/index.rst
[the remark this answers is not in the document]
JakeUrban: #195

polaris/polaris/models.py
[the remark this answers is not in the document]
JakeUrban: Actually, I'd really like all DB changes to be additive. I'm ok with not supporting this use case.

Issue #189, opened 2020-04-22
Support memo and memo_type params for deposit and withdraw endpoints


JakeUrban:
Namely, create the transaction with the memo passed. Currently this parameter is not supported.

PR #192, opened 2020-04-29
Add support for client-created memos

JakeUrban:
resolves stellar/django-polaris#189

Polaris now supports the `memo` and `memo_type` request parameters to both `/deposit/interactive` and `withdraw/interactive`.


polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -119,10 +119,9 @@ def process_response(cls, response):
     def match_transaction(cls, response: Dict, transaction: Transaction) -> bool:
         """
         Determines whether or not the given ``response`` represents the given
-        ``transaction``. Polaris does this by constructing the transaction memo
-        from the transaction ID passed in the initial withdrawal request to
-        ``/transactions/withdraw/interactive``. To be sure, we also check for
-        ``transaction``'s payment operation in ``response``.
+        ``transaction``. Polaris does this by checking the 'memo' field in the horizon
+        response matches the `transaction.memo` if present, as well as ensuring the
+        transaction includes a payment operation of the anchored asset.
```
msfeldstein, line 124: We should add some explanation of what happens if the transaction.memo isn't present.  It matches that the sender is the account authed by SEP10 instead right?  Does it require that check if the memo is present?

msfeldstein:
I think I'm good with these changes, but @tomquisel can you verify that the memo for withdraw is only needed in the case of the client being a custodial wallet?  In the case of a non-custodial wallet we can just check that the payment came from the account authed by SEP10 and don't need a memo?

polaris/polaris/management/commands/watch_transactions.py
JakeUrban: It makes sure the source of the payment op matches the stellar account on the transaction saved to the DB, so yes. And its not an either-or situation, it does both. If the memo isn't present is just compares using all the other conditions like payment amount, account, etc.

tomquisel:
### Deposit

One thing that's confusing me with Polaris is the interaction with SEP-10. In the old SEP-6, authentication wasn't always required to get deposit info or execute a withdrawal. This meant that someone with, say, a Coinbase wallet could perform a deposit of a crypto-asset and have the anchor send it to their Coinbase receive address. Does Polaris or SEP-24 allow non-authenticated access to an interactive flow? If it does not, then I don't see how a client of a custodial wallet could use Polaris. They won't be able to authenticate, as they do not control the shared account's keys.

### Withdrawal

Withdrawal is yet more confusing. The use case of the withdrawal `memo` seems to be an alternative to SEP-10. Instead of properly authenticating via SEP-10, a user can have pre-registered an account with an anchor, and could have specified their personal `memo`. They can then pass their memo into the withdrawal request so that the anchor knows which client of the custodial wallet is making a withdrawal, and can check their KYC.

So for this withdrawal memo to make sense, there needs to be an unauthenticated interactive flow, and the user needs to have indicated their memo beforehand.

Is that helpful?

JakeUrban:
Thanks for the comment @tomquisel, let me try and explain how I understand it.

> I don't see how a client of a custodial wallet could use Polaris. They won't be able to authenticate, as they do not control the shared account's keys.
A custodial wallet like Coinbase should be able to use Polaris, and actually I think this change enables that support.

**Authentication**
Coinbase (as a wallet) would authenticate its master account (the one that holds all of its users' funds) through SEP-10. I'm not sure I understand why it wouldn't be able to, since Coinbase is the owner of the account. It doesn't need the user's permission to authenticate its own account.

> Does Polaris or SEP-24 allow non-authenticated access to an interactive flow?
Regarding authentication for the interactive flow, this is done via a normal web session cookie. So Coinbase would authenticate its account, initiate the deposit (explained below), and open the interactive flow for the user to fill out. In a sense, the anchor automatically authenticates the user for the interactive flow because the wallet has already authenticated via SEP-10.

**Deposits for Custodial Accounts**
Now that Polaris supports the `memo` request parameter, custodial wallets can distinguish which user the deposit is for. The `account` would be Coinbase's master account, and the `memo` would be Coinbase's ID for the user in their own system.

When Polaris submits the deposit to the stellar network, it will now include the `memo` specified by Coinbase, so Coinbase will be able to determine which user should receive the funds.

**Withdraw for Custodial Accounts**
> The use case of the withdrawal memo seems to be an alternative to SEP-10
I don't think that would work, because someone could simply pass someone else's `memo` to the anchor. This is especially possible in SEP-6, because the `memo` is passed as a GET parameter in the unencrypted URL.

Instead, the memo serves the same purpose as it does in the deposit: to distinguish which user the funds belong to. 

Coinbase would make a request to the `/withdraw` endpoint to create a transaction request with the anchor. This request would include the master `account` parameter as well as the `memo` ID for the user. 

Once the user goes through the interactive flow (like deposit), Coinbase would send the withdrawn funds from their master account in a transaction that also has the user's ID in the memo, the same ID passed to the request. This is similar to what you were referring to by per-registering memos.

The anchor would match the memo send with the payment to the transaction created in the initial request, and send the funds to the account the user specified in the interactive flow.

Does that make sense? I know that was a lot of text.. my apologies.

JakeUrban:
We're going to close this PR because of concerns regarding custodial wallets and SEP-24's interactive flow.

PR #193, opened 2020-04-29
removed settings.ASSETS, fixed tests

JakeUrban:
resolves stellar/django-polaris#188

Adds a distribution secret key column to `Asset` and removes `settings.ASSETS`, which was previously generated from environment variable containing the asset distribution secret.

This is done to remove the need to have asset info in two separate places, DB and env vars.


polaris/polaris/settings.py
```diff
@@ -7,45 +7,16 @@
...
-        "DISTRIBUTION_ACCOUNT_ADDRESS": dist_address,
-        "ISSUER_ACCOUNT_ADDRESS": iss_address,
-    }
-
-# The SIGNING_KEY should probably be independent of the assets being anchored,
-# but if they are not specified we can use the first distribution account.
-SIGNING_SEED = env(
-    "SIGNING_SEED", default=list(ASSETS.values())[0]["DISTRIBUTION_ACCOUNT_SEED"]
-)
+SIGNING_SEED = env("SIGNING_SEED")
```
JakeUrban, line 19: This is the one potentially breaking change, `SIGNING_SEED` no longer has a default because django hasn't setup the DB before this code executes. I think its a good practice to have an explicit signing key though.

msfeldstein: agreed

docs/index.rst
```diff
@@ -115,22 +115,14 @@ environment or included in ``PROJECT_ROOT/.env``.
     DJANGO_SECRET_KEY="yoursupersecretkey"
     DJANGO_DEBUG=False
 
-    ASSETS="USD"
-    USD_DISTRIBUTION_ACCOUNT_SEED=""
-    USD_ISSUER_ACCOUNT_ADDRESS=""
-
+    SIGNING_SEED = <your signing seed here>
```
msfeldstein, line 118: Mention specifically this is for signing sep10

msfeldstein: oops i see it below nevermind

Issue #201, opened 2020-05-07
Remove need for self-signed certificate during local testing


JakeUrban:
There are a couple proposed solutions to this:

1. Don't change any code but require users to run polaris behind a proxy that supports HTTPS.

This would be a lot of work to do configure locally but one solution is to suggest users use the cloud service [ngrok](https://ngrok.com/product). This sets up a _public_ HTTPS proxy that routes traffic back to the local process.

The downside is that users have to download this tool in order to run polaris locally. They may also not want the instance to be available publicly.

2. Add a DEBUG mode that turns off interactive URL authentication, the one hard requirement for HTTPS. 

If DEBUG is true in their settings.py file, then we could not perform the functionality that requires HTTPS.

The downside is that local deployments wouldn't match the functionality in production, which could be an issue regarding testing. However if the Polaris package is tested before release there is no user-defined code that could break the session authentication.

JakeUrban:
I got the `ngrok` solution working. Here's how you'd do it:

Assuming the user has done everything mentioned in the docs up to the [Running the Server](https://django-polaris.readthedocs.io/en/latest/index.html#running-the-web-server) section:

1. Install ngrok. This is pretty easy. Download and unzip. Optionally put the binary in `/usr/local/bin`

2. Run `ngrok http --bind-tls "true" 8000

This greats the HTTPS ngrok.io subdomain and tunnel

3. Change your `HOST_URL` environment variable to the ngrok URL.

This is needed for when Polaris generates URLs for the client to request, like for the `interactive_url` response parameter or all the URLs in stellar.toml

4. Added ngrok URLs to Django's `ALLOWED_HOSTS` in settings.py:

```
ALLOWED_HOSTS = ['localhost', '127.0.0.1', '[::1]', '.ngrok.io']
```

This setting specifies which hosts the Django site can serve. This is a security measure to prevent HTTP Host header attacks. Django sets the first three items in the list above by default. Production deployments must have their hostname added to this list.

4. Run the Polaris server: `python manage.py runserver --nostatic`

---------

Overall this is probably less of a headache compared to generating a certificate, configuring your browser to trust it, and running the app with `runsslserver --certificate <path>`.

However it is still not painless. If you ngrok process is stopped, the URL changes the next time you start it, which means you have to change `HOST_URL` again.

Next I'll try the second option listed in the summary, creating a local mode where authentication for the interactive flow is disabled.

JakeUrban:
[...]

PR #202, opened 2020-05-08
Added local mode, removed sslserver dependency

JakeUrban:
resolves stellar/django-polaris#201


polaris/polaris/apps.py
```diff
@@ -9,13 +9,7 @@ def ready(self):
...
+        from polaris import settings  # ensures internal settings are set
+        from polaris.utils import check_config
 
-        if not hasattr(settings, "ACTIVE_SEPS"):
-            raise AttributeError(
-                "ACTIVE_SEPS must be defined in your django settings file."
-            )
-
-        if "sep-24" in settings.ACTIVE_SEPS:
-            check_middleware()
```
msfeldstein, line 21: why don't we check middleware anymore?

JakeUrban: We do, it was moved to `check_sep24_config()`

Issue #195, opened 2020-05-05
Explore CORS configurations


JakeUrban:
Right now Polaris requires users to accept all origins for all endpoints on the server. Ideally it would only allow all origins specifically for Polaris endpoints, and allow the user to configure the rest.

PR #203, opened 2020-05-08
Limited all-origin access to Polaris endpoints

JakeUrban:
resolves stellar/django-polaris#195


JakeUrban:
They still need to add the middleware but they don't need to add the `CORS_ALLOW_ALL_ORIGINS = True` setting

Issue #196, opened 2020-05-06
Explore solutions to the differing amounts problem


JakeUrban:
Originally discussed [here](https://github.com/stellar/django-polaris/pull/187#discussion_r420476950).

PR #205, opened 2020-05-12
Allow differing amounts in watch_transactions

JakeUrban:
resolves stellar/django-polaris#196

Before this change, funds sent to an anchor's distribution account had to exactly match the amount specified by the user during the interactive flow. This change removes that requirement and now allows the anchor to decide how to handle differing amounts.

Since transactions submitted to Horizon and `Transaction` DB objects in Polaris can no longer be matched using the amount, `memo` is now the only deterministic way to find a match. 

On both SEP 6 and 24, the `memo` request parameter for `/withdraw` and the `memo` attribute returned by the `/transaction` endpoint have different uses. `memo` sent in the request should be used by the anchor to disambiguate users who use the same stellar account, whereas `memo` returned in the `/transaction` endpoint response should be used by the wallet when submitting the transaction.

So, this PR also ensures that both SEP 6 and SEP24 withdrawal transactions have memo's created by the anchor in order to later match them with transactions sent over the network by the wallet. Previously, SEP-6 would use the `memo` sent in the `/withdraw` request and leave it `null` if one was not sent.

This PR also currently returns 400 for memo's sent by clients to SEP-24 `/deposit` and `/withdraw` endpoints, since we do not want to give clients the impression that Polaris SEP-24 supports custodial wallet withdraws. This policy may change.

### DIFF
--- a/polaris/polaris/integrations/transactions.py
+++ b/polaris/polaris/integrations/transactions.py
@@ -317,9 +317,25 @@ class WithdrawalIntegration:
         """
         .. _endpoint: https://www.stellar.org/developers/horizon/reference/resources/transaction.html
 
-        This method should implement the transfer of the amount of the
-        anchored asset specified by `transaction` to the user who requested
-        the withdrawal.
+        This method is called when the transacted asset's distribution account receives
+        a payment from a transaction with a memo matching `transaction.memo`.
+
+        If `transaction` was created via SEP-24, it is very important to confirm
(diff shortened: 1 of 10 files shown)


JakeUrban:
@msfeldstein lets hold off on this until we come to an agreement on use of memos in SEP-24

JakeUrban:
@msfeldstein Ok this is ready 👍

Issue #165, opened 2020-03-05
Feature Request: Add fee information to the interactive flow


vitchor:
Show how many SRTs the user will actually receive (taking into account the fees) in the deposit and withdraw amount screens.

JakeUrban:
I have the functionality implemented, but where should the fee information be displayed?
![Screen Shot 2020-05-14 at 9 56 32 AM](https://user-images.githubusercontent.com/10968980/81962938-4c57a980-95c9-11ea-8117-1e26f4d15e76.png)

Under the submit button? What should the color and font size be of the text?

JakeUrban:
I create a separate page that displays this information, like a "confirm" page where they see the amount in, fee, and amount out.

PR #211, opened 2020-05-15
Added Fee info to TransactionForm HTML

JakeUrban:
resolves stellar/django-polaris#165

`/fee` endpoint no longer requires authentication, this was necessary to hit the endpoint from the client without the SEP10 JWT.

Added `Asset.symbol` so each anchor can display the correct currency symbol in front of their amounts.

Added JS and HTML for displaying the fee and total for the amount deposited/withdrawn in real time.


JakeUrban:
TODO: add translations

polaris/polaris/tests/shared_endpoints/test_fee.py
```diff
@@ -201,6 +201,4 @@ def test_fee_no_jwt(client, usd_asset_factory):
     response = client.get(
         f"{endpoint}?asset_code=USD&operation=withdraw&amount=100.0", follow=True
     )
-    content = json.loads(response.content)
-    assert response.status_code == 403
-    assert content == {"error": "JWT must be passed as 'Authorization' header"}
+    assert response.status_code == 200
```
msfeldstein, line 204: I think we can just remove this whole test, or at least update the name/comment

PR #212, opened 2020-05-18
Disable secure session cookie when in local mode

JakeUrban:
Our custom middleware was adding the `Secure` flag to session cookie thats created when you login to the admin panel. This PR disables use of that flag when in local mode.

Side note: Do we want to continue using PolarisSameSiteMiddleware? This was a fix for iframes, which we no longer recommend in the SEP.


msfeldstein:
we don't disallow iframes so we may as well keep it for now

Issue #214, opened 2020-05-26
Add SEP31 to Polaris


msfeldstein:
Top-level ticket for implementing SEP-31 in Polaris

- [x] Make necessary DB migrations
- [x] Update admin menus
- [x] Update `watch_transactions` CLI tool for SEP31
- [x] Add `poll_outgoing_transactions` CLI tool
- [x] Add `execute_outgoing_transactions` CLI tool
- [x] Implement `/send` & associated integration
- [x] Implement `/update` & associated integration
- [x] Implement `/info` & associated integration
- [x] Implement `/transaction` & SEP31 transaction serializer
- [x] Implement callback functionality
- [x] Write tests for all the above
- [x] Add docstrings and documentation
- [x] Implement SEP31 support in example reference server

PR #218, opened 2020-06-15
SEP31 Support

JakeUrban:
resolves stellar/django-polaris#214

Adds SEP31 Support to Polaris

**Endpoints & Integrations**

Implements all SEP31 endpoints. Introduces the `SendIntegration` class, which has the following methods:

- `info`
- `process_send_request`
- `process_update_request`
- `valid_sending_anchor`
    - returns `True` or `False` for a given stellar address

Also introduces the `RailsIntegration` class. All rails-related functions will eventually be moved here, including those from existing integration classes. For now, `RailsIntegration` has two integrations:

- `execute_outgoing_transaction`
    - executes the off-chain payment to the user
- `poll_outgoing_transactions`
    - checks the statuses of off-chain payments and marks them as `completed` when the user has received the funds.

So the flow is
1. Sending anchor makes `/send` request
2. Sending anchor makes stellar transaction to receiving anchor
3. Receiving anchor's `watch_transactions` matches the payment to a pending transaction
4. `watch_transactions` marks the transaction as `pending_receiver`
5. `execute_outgoing_transactions` (new process) executes the off-chain payment to the user
6. `poll_outgoing_transactions` (new process) marks the payment as `completed` when the user has received the funds.

**DB Changes**

Transaction
- added `Transaction.PROTOCOL.sep31` value
- added `Transaction.KIND.send` value
- added SEP31 statuses (`pending_sending`, `pending_receiver`, etc.)
- renamed `external_extra` and `external_extra_text` to `required_info_update` and `required_info_message`, respectively. The former 2 were not used.
- added `send_memo`, `send_memo_type`, `send_fee_fixed`, `send_fee_percent`, `send_max_amount`, `send_min_amount`, `send_anchor_account`, and `sep31_enabled` columns

**Processes**

Added two additional processes to run alongside the web server:

- `execute_outgoing_transactions`: This process continuously checks for transactions that are ready to be deposited to the receiving user's off-chain account, and executes the deposits.
- `poll_outgoing_transactions`: This process continuously checks the off-chain deposits made to the receiving user and updates their `status` to `completed` once the user has received the funds.

**Future Changes**

All existing integration functions that connect to off-chain rails will be moved to `RailsIntegration`. This makes sense because many SEPs require common rail-related functions, such as sending money to users, either for withdraws (SEP24/6) or payments (SEP31).

The functions that will be moved:
- `poll_pending_deposits` --> `poll_incoming_transactions`
    - checking for incoming transfers to an anchor's off-chain account is SEP-independent
- `process_withdraw`
    - this will likely be merged with `execute_outgoing_transaction` since they perform the same function

Additionally, `register_integrations` currently accepts an `info_func` keyword argument. This should be changed to `sep6_info_func` since SEP31 also has an `info` integration inside `SendIntegration`.


polaris/polaris/management/commands/execute_outgoing_transactions.py
```diff
@@ -0,0 +1,90 @@
...
+            self.execute_outgoing_transactions()
+
+    @staticmethod
+    def execute_outgoing_transactions():
+        # For the time being this function is only for SEP 31 transactions
+        # Eventually we'll process transfers for SEP 6 and SEP 24 transactions
+        # here as well.
+        transactions = Transaction.objects.filter(
+            protocol=Transaction.PROTOCOL.sep31,
+            status=Transaction.STATUS.pending_receiver,
```
msfeldstein, line 47: is there any other reason a transaction might be pending_receiver? I don't think so but just want to make sure

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -82,7 +82,7 @@ def execute_deposits(cls):
             # every time its called, but we're going to assume it was a special
             # case and allow the process to continue running by returning instead
             # of re-raising the error. The anchor should see the log messages and
-            # fix the issue if it is reoccuring.
+            # fix the issue if it is reoccurring.
```
msfeldstein, line 85: frankly im shocked that there is two cc's in reoccurring.

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -140,9 +206,15 @@ def find_matching_payment_op(
         except KeyError:
             return
 
-        # memo from response must match transaction.memo
+        # memo from response must match transaction memo
         memo = response.get("memo")
-        if memo != transaction.withdraw_memo:
+        if (
```
msfeldstein, line 211: This should mean that its sep6 right?  Why not `transaction.protocol == Transaction.PROTOCOL.sep6`?

polaris/polaris/models.py
```diff
@@ -305,9 +349,33 @@ class Transaction(models.Model):
     * **error**
     
         catch-all for any error not enumerated above.
+        
+    SEP-31 Statuses:
+    
+    * **pending_sender**
+    
+        awaiting payment to be initiated by sending anchor.
```
msfeldstein, line 357: Should the whitespace be under this line instead of above?

example/server/models.py
```diff
@@ -47,8 +47,15 @@ class PolarisUserTransaction(models.Model):
     this table serves to join the two entities.
     """
 
-    transaction = models.OneToOneField(Transaction, on_delete=models.CASCADE)
-    account = models.ForeignKey(PolarisStellarAccount, on_delete=models.CASCADE)
+    transaction_id = models.TextField(db_index=True)
```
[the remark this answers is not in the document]
JakeUrban: At the time of calling the `process_send_request()` integration, the `Transaction` object isn't created yet, since the integration could return a 400 response.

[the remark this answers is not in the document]
JakeUrban: So our reference anchor using Polaris saves the ID that _will_ be used to create the transaction, but not in a FK column which would check for the row's existence when inserted.

example/server/integrations.py
```diff
@@ -450,6 +454,143 @@ def delete(self, account: str):
...
+            PolarisUserTransaction.objects.filter(transaction_id=transaction.id)
+            .first()
+            .user
+        )
+
+        client = rails.BankAPIClient("fake anchor bank account number")
+        transaction.amount_fee = 0  # Or calculate your fee
+        response = client.send_funds(
+            to_account=user.bank_account_number,
+            amount=transaction.amount_in - transaction.amount_fee,
```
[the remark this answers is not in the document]
JakeUrban: The amount received by the sending anchor should be in the receiving user's currency units, so it should be.

polaris/polaris/management/commands/execute_outgoing_transactions.py
JakeUrban: Nope, I think thats it. But thats a big assumption though. I can't think of another reason.

example/server/integrations.py
```diff
@@ -450,6 +454,143 @@ def delete(self, account: str):
...
+                bank_account_number=transaction["account_number"],
+                bank_number=transaction["routing_number"],
+            )
+        elif not (user.bank_account_number and user.bank_number):
+            user.bank_account_number = transaction["account_number"]
+            user.bank_number = transaction["routing_number"]
+            user.save()
+        # Transaction doesn't yet exist so transaction_id is a text field
+        PolarisUserTransaction.objects.create(user=user, transaction_id=transaction_id)
+        # Don't handle receiver_info yet
```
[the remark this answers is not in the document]
JakeUrban: Actually I meant `receiver_info`, assuming the wallet sends `require_receiver_info`. But you're right, we don't actually use sender data either.

example/server/integrations.py
```diff
@@ -450,6 +454,143 @@ def delete(self, account: str):
...
+
+    def process_update_request(self, params: Dict, transaction: Transaction):
+        info_fields = params.get("fields", {})
+        receiver_fields = info_fields.get("receiver", {})
+        transaction_fields = info_fields.get("transaction", {})
+        possible_fields = [
+            "first_name",
+            "last_name",
+            "routing_number",
```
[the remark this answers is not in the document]
JakeUrban: Yea I can refactor this.

polaris/polaris/management/commands/watch_transactions.py
JakeUrban: Nope. This condition is to make sure we're comparing `memo` with the right column on `transaction`. We could do `transaction.protocol in [Transaction.PROTOCOL.sep6, Transaction.PROTOCOL.sep24]` OR `transaction.protocol != Transaction.PROTOCOL.sep31`

polaris/polaris/models.py
JakeUrban: It looks like it in the code but when its rendered, this is right.

polaris/polaris/sep31/info.py
```diff
@@ -0,0 +1,92 @@
...
+@renderer_classes([JSONRenderer, BrowsableAPIRenderer])
+def info(request: Request) -> Response:
+    info_data = {
+        "receive": {},
+    }
+    for asset in Asset.objects.filter(sep31_enabled=True):
+        try:
+            fields = registered_send_integration.info(asset, request.GET.get("lang"))
+        except ValueError:
+            return render_error_response("unsupported 'lang'", content_type="text/html")
```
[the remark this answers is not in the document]
JakeUrban: This is the integration function, so if the developer using Polaris throws an error from `info()` it will be caught here. Polaris assumes the error is due not being able to support the language, since there is no other reason the `/info` endpoint would ever return a 400 response.

Issue #222, opened 2020-07-02
Update SEP-12


JakeUrban:
SEP-12 has [incoming changes](https://github.com/stellar/stellar-protocol/pull/653) that will allow it to be used by other SEPs, like SEP-31.

PR #223, opened 2020-07-02
Added SEP-12 GET method, fixed DELETE

JakeUrban:
resolves stellar/django-polaris#222

Adds `GET /customer` to Polaris' SEP-12 support and adds an `id` response attribute to `PUT /customer`. Adds tests for all endpoint methods.

Added `memo` and `memo_type` columns to reference servers' `PolarisStellarAccount` model since SEP-12 requests for SEP-31 transactions will use memos.

Also fixed a bug for the `DELETE /customer` request. In previous versions this does endpoint does not work. These releases (0.11.0, 0.11.1) will be removed from PyPI when 0.11.2 is added.


example/server/integrations.py
```diff
@@ -388,7 +390,69 @@ def process_sep6_request(self, params: Dict) -> Dict:
...
+                    "last_name": {"description": "last name of the customer"},
+                    "email_address": {"description": "email address of the customer"},
+                },
+            }
+            if params.get("type") in ["sep6-deposit", "sep31-sender"]:
+                return response
+            elif params.get("type") in [None, "sep6-withdraw", "sep31-receiver"]:
+                response["fields"].update(
+                    {
+                        "bank_account_number": {
```
msfeldstein, line 419: All this logic feels kind of messy, it'd be nice to have a simple mapping of what fields are needed by what types, and then the logic can just iterate through that rather than put the needed fields in all the if statements

example/server/integrations.py
```diff
@@ -400,13 +464,16 @@ def put(self, params: Dict):
         if not all(val in params for val in required_fields):
             raise ValueError(f"required fields: {', '.join(required_fields)}")
 
+        # query params for fetching/creating the PolarisStellarAccount
+        qparams = {"account": params["account"]}
+        if "memo" in params:
+            qparams["memo"] = params["memo"]
+            qparams["memo_type"] = params["memo_type"]
```
msfeldstein, line 471: Possible to pass a memo with no memotype?

polaris/polaris/sep12/customer.py
```diff
@@ -1,40 +1,106 @@
...
-    else:
-        return Response({}, status=202)
+logger = Logger(__name__)
+
+
+class CustomerAPIView(APIView):
+    renderer_classes = [JSONRenderer]
+
+    @staticmethod
+    @validate_sep10_token("sep6")
```
msfeldstein, line 22: Why are we referencing sep6 in a sep12 file?

example/server/integrations.py
JakeUrban: Nope, `memo_str` will raise an error in that case

polaris/polaris/sep12/customer.py
JakeUrban: SEP-12 is used in SEP-6, and SEP-6 has all 403 permission denied responses to have this body:
```
{
    "type": "authentication_required"
}
```
So I added the `"sep6"` string to ensure thats the response body on authenticated requests. 

Since SEP-31 isn't opinionated about the response body of 403 responses it doesn't conflict.

PR #230, opened 2020-07-13
Breaking changes: Rails Integrations

JakeUrban:
This PR makes breaking changes to the API provided by Polaris' integration classes and functions

**Rails-related Functions:**

First, I moved `DepositIntegration.poll_pending_deposits` to `RailsIntegration.poll_pending_deposits`. I'm considering renaming this function to `poll_incoming_transactions` to match the newly added process `poll_outgoing_transactions`.

The reason to move this function is simple. Not every deposit into the anchor's external accounts are for SEP-6 or SEP-24. For example, a SEP-31 sending anchor can expect users to send them money so they can forward those funds to a receiver. Polaris does not support being a SEP-31 sending anchor, but it could in the future. The point is that SEP-6 and SEP-24 don't have a monopoly on an anchor's incoming non-stellar transactions, and all rails-related functions should be on `RailsIntegration` anyway.

Secondly, I removed `WithdrawalIntegration.process_withdrawal` and added support for SEP-6 and SEP-24 withdraw transactions going through `execute_outgoing_transactions()`, which is currently a SEP-31-only integration. The expectation is that the anchor transfers the funds from their stellar account to the user's off-chain account.

These two functions, `process_withdrawal()` and `execute_outgoing_transactions()`, serve the same purpose. The difference is that `process_withdrawal()` was, like `poll_pending_deposits()`, tied to the SEP-6 and SEP-24 integration classes. Since transferring assets off-chain is not exclusive to SEP-6 and SEP-24, this change needed to be made.

### DIFF
--- a/example/server/integrations.py
+++ b/example/server/integrations.py
@@ -1,5 +1,3 @@
-from typing import Union
-
 import json
 from smtplib import SMTPException
 from decimal import Decimal
@@ -101,7 +99,7 @@ class SEP24KYC:
                 )
 
         PolarisUserTransaction.objects.get_or_create(
-            account=account, transaction_id=transaction.id
+            user=account.user, account=account, transaction_id=transaction.id
         )
 
     @staticmethod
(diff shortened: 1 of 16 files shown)


JakeUrban:
I removed the fee integration changes because it makes sense as a separate PR.

example/server/integrations.py
```diff
@@ -586,7 +545,7 @@ def info(self, asset: Asset, lang: Optional[str] = None):
         }
 
     def process_send_request(self, params: Dict, transaction_id: str) -> Optional[Dict]:
-        sender_id = params.get("sender_id")  # not actually used
+        _ = params.get("sender_id")  # not actually used
```
msfeldstein, line 548: whats the purpose of this line?

JakeUrban: `_` is a way of not assigning a variable. I just have the line to show that we collect `sender_id`. We don't use the sender info when processing both those looking at the code for reference might.

Issue #184, opened 2020-04-17
Use ORG_LOGO for icon in interactive flow


JakeUrban:
Polaris currently requires anchors to add an image file to their static assets directory in order to display their logo on the interactive flow. Instead, the templates could retrieve the image from the stellar.toml's ORG_LOGO link.

Issue #233, opened 2020-07-16
Simplify content_for_transaction() return type


JakeUrban:
Right now `content_for_transaction()` returns a dictionary:
```
{
    "form": form,  # initialized django form
    "icon_label": "Stellar Development Foundation",
    "guidance": "Enter the amount you would like to deposit",
    "title": "SEP-24 Deposit"
}
```

I'd like to simply return the `form` object. My current solution is to have Polaris expect attributes on the `form` object for `icon_label`, `guidance`, and `title`. But the anchor may want to have these things on a page without a form.

We need a way for anchors to specify the label, guidance, and title per-page without tying it to the form. For example, currently the more_info.html page doesn't have an icon label because there is no form.

Then we can simply `content_for_transaction()` which may be renamed to `form_for_transaction()`

We could add an integration function `def content_for_template(template_name, form=None, transaction=None)`. Functionally doing 
```
form = form_for_transaction(transaction)
content = content_for_template("deposit.html", form, transaction)
```
Is equivalent to
```
content = content_for_transaction(transaction)
```

But separating the `form` from the page content allows Polaris to ask for content for a page without a form and simplifies the original integration function.

PR #239, opened 2020-07-20
Allow users to specify path to an image in their static directory OR use ORG_LOGO

JakeUrban:
resolves stellar/django-polaris#184

Currently this isn't a full solution, since the more info page doesn't use `content_for_transaction()`. But I have another ticket filed to split that function into two, so the more info page will use `content_for_template()`, which will have the `icon_label` key-value pair in its return type (See #233)

### DIFF
--- a/example/server/integrations.py
+++ b/example/server/integrations.py
@@ -126,7 +126,6 @@ class SEP24KYC:
                         "Please enter the information requested."
                     )
                 ),
-                "icon_path": "images/lobstr.png",
             }
         elif settings.LOCAL_MODE:
             # When in local mode, request session's are not authenticated,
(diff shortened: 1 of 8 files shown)


example/server/integrations.py
```diff
@@ -140,6 +141,7 @@ def check_kyc(transaction: Transaction, post_data=None) -> Optional[Dict]:
                     "continue on this page."
                 ),
                 "icon_label": _("Stellar Development Foundation"),
+                "icon_path": "images/lobstr.png",
```
msfeldstein, line 144: i don't think we should use a third party companies logo in our example

JakeUrban: Oh yea that was a test. We don't have a PNG file to use, so we'll just use the Polaris default.

Issue #252, opened 2020-07-27
SEP-10 not HTTP compliant


shredding:
This does not work well with W3s recommendation of always adding a charset: https://github.com/stellar/django-polaris/blob/master/polaris/polaris/sep10/views.py#L186-L193

See: https://www.w3.org/International/articles/http-charset/index.en#charset

e.g. `application/json, charset: utf-8` breaks, hence client libraries that follow best practises needs to be hacked.

I'm not sure why that decision was made, but unless there are very good reasons not to do so DRF ships with HTTP-compliant Content Negotation out of the box, and just adding

```
parser_classes = [FormParser, JSONParser]
```

to the APIView would allow to get rid of the entire self-written content negotatiation.

JakeUrban:
I agree, good catch. It will be fixed in the next release.

PR #258, opened 2020-07-27
Don't try to manually decode SEP-10 request body

JakeUrban:
resolves stellar/django-polaris#252

This one is pretty straight forward. The original skeleton code has the functions removed here, I never touched them since they worked. But according to the issue, it actually didn't work all the time.

### DIFF
--- a/polaris/polaris/sep10/views.py
+++ b/polaris/polaris/sep10/views.py
@@ -10,10 +10,11 @@ import binascii
 import time
 import jwt
 
-from django.http import JsonResponse
+from django.utils.translation import gettext as _
 from rest_framework import status
 from rest_framework.views import APIView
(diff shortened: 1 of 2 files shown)


shredding:
In order to maintain exisiting applications, you should still `raise ValueError("no transaction found")` if `envelope_xdr` is None in line 88 (read_challenge_transaction breaks in that case as well).

JakeUrban:
Agreed, there are more error cases I need to catch as well. I rushed this into a PR and it should probably be considered a draft.

JakeUrban:
Filed an issue to add SEP-10 unit tests. For now I can rely on the SDF's anchor-validator.stellar.org test suite. Looks like its passing everything now.

Issue #247, opened 2020-07-23
SEP-6 Link created Transaction object in process_sep6_request()


yuriescl:
In SEP-6 there can be custom fields defined by the Anchor:
```
GET /info

{
  "deposit": {
    ...
    "fields": {
      "email_address": {
        "description": "Beneficiary email address",
        "optional": False,
      },
      "wallet_name": {
        "description": "Wallet name",
        "optional": True,
      },
    }
  }
}
```

When SEP-6 `/deposit` or `/withdraw` is called, to handle the validation, I imagine I'd have to implement and register the function `DepositIntegration.process_sep6_request`.
Besides validating the values, I'd like to save them and associate them with the Transaction created in SEP-6, maybe using a new model `TransactionExtra` and having a `ForeignKey` pointing to the SEP-6  `Transaction`.

My issue is that Polaris creates the Transaction after calling `process_sep6_request`, so there's no way for me to create a `TransactionExtra` because there would not be a `Transaction` to point to yet.
I'd like to know if there's any recommendation regarding saving the values of these fields and associating them with the `Transaction`.

JakeUrban:
The reason `Transaction` is not created before the call to `process_sep6_request` is because the function could return a 403 response, in which case Polaris shouldn't create a Transaction.

However, this obviously is inconvenient for developers using Polaris. The same problem arose while implementing SEP-31 support, so I added a `transaction_id` parameter to `process_send_request` (the SEP-31 equivalent of `process_sep6_request`). 

The `Transaction` record in the DB still doesn't exist when `process_send_request` is called, but it still allows the developer to save the ID and use it to connect the transaction to other tables. This is better but still not ideal.

We're going to release 1.0 soon and are currently making some breaking changes to the API. One change we could make is to pass a `Transaction` object to `process_sep6_request` and `process_send_request`, and allow the developer to save the `Transaction` object to the DB if they plan to return a success response.

Another solution could be adding post-transaction-creation integration functions to the respective SEP-6 and SEP-31 endpoints. This would allow you to use the transaction after its saved to the DB. I'll file an issue for this discussion.

While I work on a new release with one of the above solutions ^, I can think of a solution, although not ideal. Have you heard of [Django Signals](https://docs.djangoproject.com/en/2.2/topics/signals/)? You can register a django signal function for the `Transaction` `post_save` event and connect it to the rest of your data model if the transaction was just created.

yuriescl:
[...]

PR #261, opened 2020-07-28
Pass transaction to process_sep6_request integration functions

JakeUrban:
resolves stellar/django-polaris#247

As described in the issue, this change passes the `Transaction` object to `process_sep6_request()`, which allows the developer to link the transaction created as a result of the request to the rest of the anchor's data model.

I'll also be updating SEP-31's `process_send_request()` integration function to pass the `Transaction` object instead of the id to match this solution.


JakeUrban:
@yuriescl

PR #277, opened 2020-08-07
change script integration to file names

acharb:
**WHY**

Makes it simpler for the user to integrate js files, instead of putting js code in a python file

### DIFF
--- a/example/server/integrations.py
+++ b/example/server/integrations.py
@@ -25,6 +25,7 @@ from polaris.integrations import (
     calculate_fee,
     RailsIntegration,
     TransactionForm,
+    TemplateScript
 )
 from polaris import settings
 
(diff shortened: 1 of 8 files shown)


example/server/integrations.py
```diff
@@ -783,77 +783,82 @@ def toml_integration():
...
+    #     <script async src="https://www.googletagmanager.com/gtag/js?id=UA-53373928-6"></script>
+    #     <script>
+    #       window.dataLayer = window.dataLayer || [];
+    #       function gtag(){dataLayer.push(arguments);}
+    #       gtag('js', new Date());
+    #       gtag('config', 'UA-53373928-6');
+    #     </script>
+    #     """
+    # ]
+    # if (
```
JakeUrban, line 803: This logic still needs to be present in the integration function. We don't want to load this script unless its a particular page being served.

example/server/static/sep24_scripts/google_analytics.js
```diff
@@ -0,0 +1,6 @@
+window.dataLayer = window.dataLayer || [];
```
JakeUrban, line 1: What is this? If this was just a test can we delete it?

example/server/static/sep24_scripts/test.js
```diff
@@ -0,0 +1,5 @@
+function x() {
```
JakeUrban, line 1: Same here.

polaris/polaris/templates/base.html
```diff
@@ -222,8 +222,8 @@
     </script>
 
     {# Rendered via Polaris javascript integrations #}
-    {% for script in scripts %}
-      {{ script|safe }}
+    {% for script_src in scripts %}
+      <script src="/static/{{ script_src }}"></script>
```
JakeUrban, line 226: The template logic will get more complex if we want to allow anchors to choose if the script should load synchronously or asynchronously. You'll likely have to iterate through a list of `TemplateScript` objects and check for an `async` property.

example/server/integrations.py
```diff
@@ -783,77 +783,82 @@ def toml_integration():
 
 
 def scripts_integration(page_content: Optional[Dict]):
```
JakeUrban, line 785: If we're going with the async/sync per-script functionality, we need to adjust the return type. Make sure you update the `scripts()` default javascript function with the new return type. The return type should be able to communicate the async/sync thing for now, but allow for future attributes to be added. (I'm thinking an object)

example/server/static/sep24_scripts/google_analytics.js
acharb: This was there from before as part of that google_analytics script

polaris/polaris/integrations/javascript.py
```diff
@@ -1,21 +1,40 @@
 from typing import List, Dict, Optional
 
 
-def scripts(page_content: Optional[Dict]) -> List[str]:
+class TemplateScript:
+    def __init__(self, file_path: str = None, url: str = None, is_async: bool = False):
```
JakeUrban, line 5: Can we change `file_path` to `path` and `is_async` to `async`?

polaris/polaris/integrations/javascript.py
```diff
@@ -1,21 +1,40 @@
...
+        if file_path and url:
+            raise AttributeError("A script can only have one source from either a file_path or url.")
+        elif not (file_path or url):
+            raise AttributeError("Must give a source for TemplateScript.")
```
JakeUrban, line 9: I would explicitly state what parameters are necessary here. Also start the error messages with a lower-case letter.

example/server/integrations.py
```diff
@@ -783,19 +784,13 @@ def toml_integration():
...
-        <script async src="https://www.googletagmanager.com/gtag/js?id=UA-53373928-6"></script>
-        <script>
-          window.dataLayer = window.dataLayer || [];
-          function gtag(){dataLayer.push(arguments);}
-          gtag('js', new Date());
-          gtag('config', 'UA-53373928-6');
-        </script>
-        """
-    ]
+    tag_paths = []
```
JakeUrban, line 787: Can we change `tag_paths` to `scripts` or something more representative? And initialize the list with the contents:
```
scripts = [
    # Google Analytics
    # <!-- Global site tag (gtag.js) - Google Analytics -->
    tag_paths.append(TemplateScript(url="https://www.googletagmanager.com/gtag/js?id=UA-53373928-6", is_async=True))
    tag_paths.append(TemplateScript(file_path="sep24_scripts/google_analytics.js"))    
]
```

polaris/polaris/integrations/javascript.py
acharb: I can change to path, async is reserved by python so throws an error when you try to use as an argument name. 

I thought is_async would be better than is_sync, since async is the attribute getting added to the html

JakeUrban: Oh right ok no worries then, `is_async` is alright

Issue #266, opened 2020-07-29
Add testnet reset CLI tool


JakeUrban:
There are several DB operations necessary for testnet anchors every time the network resets.

- The most recent `Transaction` completed for each anchored asset's distribution needs its `paging_token` set to `None`
- All `pending_trust` `Transaction` objects must be updated to `error` since the account needed a trustline was deleted

I'll add more if there is anything else.

PR #278, opened 2020-08-11
Testnet CLI command

JakeUrban:
resolves stellar/django-polaris#266

Adds a Django command so anchors can issue assets mirroring the https://github.com/stellar/create-stellar-token project.

Also adds a `reset` subcommand for reseting the most-recent `paging_token` for each anchored asset and re-issues each asset in the DB. `paging_token` must be reset when testnet resets so Polaris can begin streaming from the network again.

Also includes formatting changes from @acharb 's recent commit on script integrations. Polaris uses the [black](https://pypi.org/project/black/) formatter for its python code.

TODO: update `pending_trust` transactions to `error`

### DIFF
--- a/polaris/polaris/management/commands/testnet.py
+++ b/polaris/polaris/management/commands/testnet.py
@@ -69,6 +69,7 @@ class Command(BaseCommand):
         state after a testnet reset. Currently this involves the following:
 
         - re-issuing every Asset object in the DB
+        - moves all `pending_trust` Transactions to `error`
         - setting the most-recently streamed Transaction object's
             `paging_token` attribute to None. This signals to
             watch_transactions to stream using the `"now"` keyword.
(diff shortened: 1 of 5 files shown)


polaris/polaris/management/commands/testnet.py
```diff
@@ -0,0 +1,225 @@
...
+        state after a testnet reset. Currently this involves the following:
+
+        - re-issuing every Asset object in the DB
+        - setting the most-recently streamed Transaction object's
+            `paging_token` attribute to None. This signals to
+            watch_transactions to stream using the `"now"` keyword.
+        """
+        print("\nResetting each asset's most recent paging token")
+        for asset in Asset.objects.filter(distribution_seed__isnull=False):
+            Transaction.objects.filter(
```
JakeUrban, line 78: Get only the first TX, not all.

Issue #243, opened 2020-07-22
Squash migration files


JakeUrban:
Polaris has accumulated a significant amount of DB changes that come with migration files. We should reset the migration tree and squash all changes up to this point before the 1.0 relase.

https://docs.djangoproject.com/en/3.0/topics/migrations/#squashing-migrations

PR #281, opened 2020-08-14
Squashed migrations

JakeUrban:
resolves stellar/django-polaris#243

There are no new migrations from 0.13.1 to 1.0. Since we'll be releasing the first stable version, its a good time to reset Polaris'  migrations. 

Those upgrading from older versions of the package should upgrade to 0.13.1 first, which is the last version with the pre-1.0 migrations. Everyone will have to ensure their migrations DB table is updated to reflect the `migrations` polaris directory.


polaris/polaris/migrations/0001_initial.py
```diff
@@ -43,33 +44,132 @@ class Migration(migrations.Migration):
                         verbose_name="modified",
                     ),
                 ),
+                ("code", models.TextField(default="USD")),
```
JakeUrban, line 47: Don't merge this, we should remove the default

Issue #248, opened 2020-07-26
Feature Request: get settings from settings.py if not in .env


yuriescl:
Polaris currently requires the Django app to create a `.env` file and put the Polaris settings there.
But Django settings are usually placed in `settings.py`, possibly overriden in a `local_settings.py`.
If the developer wants to get the settings from the environment (`.env` or environment variables), he should have the option to do so, but that should not be enforced like it currently is by Polaris.

My suggestion is to either remove the need for a `.env` file, or make it optional to get the settings from the environment and maybe control that option through a flag in `settings.py`, something like `POLARIS_ENV`.

The main reason for asking this is because I'd like to have all settings on `settings.py`, with local settings in `local_settings.py`, this way I can avoid multiple settings files (.json, .yaml, etc).
Celery for example reads settings from `settings.py` by default, with the prefix `CELERY_`.
So Polaris could have something like:
```
# settings.py

POLARIS_STELLAR_NETWORK_PASSPHRASE = ""
POLARIS_SIGNING_SEED = ""
```

yuriescl:
Also, it's strange that currently some settings are required to be in `settings.py` and others in `.env`.
Currently these settings are required by Polaris in `settings.py`:
```
ACTIVE_SEPS = ['sep-6', 'sep-10']
SECURE_SSL_REDIRECT = True
SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
```
If some settings can be in `settings.py`, why not all settings?
```
POLARIS_STELLAR_NETWORK_PASSPHRASE = ""
POLARIS_SIGNING_SEED = ""
...
POLARIS_ACTIVE_SEPS = ['sep-6', 'sep-10']

SECURE_SSL_REDIRECT = True
SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
```

JakeUrban:
Before I address the issue, I actually really like the idea of prepending polaris-specific settings in `settings.py` with `POLARIS_`. I think I'll add that for the next release.

[...]

PR #283, opened 2020-08-14
Use settings.py if variable is not found in environment

JakeUrban:
resolves stellar/django-polaris#248

Looks in `django.conf.settings` (settings.py) if no variable with the given name is found in the environment. This allows Polaris users to put all settings in settings.py.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -135,7 +135,7 @@ Polaris also supports specifying your environment variables in your project's ``
     POLARIS_HORIZON_URI = "https://horizon-testnet.stellar.org/"
     POLARIS_HOST_URL = "https://example.com"
 
-If a variable cannot be found in the environment, Polaris will look in the ``settings.py`` file. Be careful not to check in secrets like your ``SERVER_JWT_KEY`` into your version control history, especially if your repository is not private.
+If a variable cannot be found in the environment, Polaris will look in the ``settings.py`` file, or more generally, ``django.conf.settings``. Be careful not to check in secrets like your ``SERVER_JWT_KEY`` into your version control history, especially if your repository is not private.
 
 Endpoints
 ^^^^^^^^^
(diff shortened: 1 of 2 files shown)


JakeUrban:
@yuriescl

docs/index.rst
```diff
@@ -121,14 +121,22 @@ You may want to configure the ``LEVEL`` of the Polaris logger differently depend
...
     HOST_URL="https://example.com"
 
+Polaris also supports specifying your environment variables in your project's ``settings.py``. However, any variable Polaris expects in the environment must be prepended with ``POLARIS_`` if declared in `settings.py``. For example,
+::
+
+    POLARIS_STELLAR_NETWORK_PASSPHRASE = "Test SDF Network ; September 2015"
+    POLARIS_HORIZON_URI = "https://horizon-testnet.stellar.org/"
+    POLARIS_HOST_URL = "https://example.com"
+
+If a variable cannot be found in the environment, Polaris will look in the ``settings.py`` file. Be careful not to check in secrets like your ``SERVER_JWT_KEY`` into your version control history, especially if your repository is not private.
```
yuriescl, line 138: @JakeUrban This makes it sound like Polaris will look in the `settings.py` file directly. I suggest mentioning that Polaris will look for settings in `django.conf` and `settings.py` is where these settings are usually defined, but also mention that `local_settings.py` should be used to store secret settings:
```
# bottom of settings.py
try:
    from local_settings import *
except ImportError:
    pass
```

JakeUrban: I can mention that Polaris reads specifically from `django.conf` instead of an actual `settings.py` file, but I think the `local_settings` module you use doesn't need to be mentioned in the docs. 

Keeping python files secret is a bit unusual, as opposed to secret environment variables. If Polaris users are declaring all their variables in python, they're likely declaring them explicitly in `settings.py` and keeping the entire repository secret/private. Even if they use your approach, its implied that other modules can be imported into the main `settings.py`.

JakeUrban:
@yuriescl if you could test the code and ensure it works with your `local_settings.py` configuration that would be great.

yuriescl:
@JakeUrban It's working, thank you.

Issue #236, opened 2020-07-16
Upgrade to Django 3.1 (August 2020)


JakeUrban:
As long as the packages we use alongside django work with Django 3, upgrade to it when its released.

This will allow us to remove `PolarisSameSiteMiddleware` from the setup steps.

We may be able to remove this step already since we use popups not iframes, as stated in [SEP-24](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#basic-anchor-implementation).

JakeUrban:
So the main blocker for upgrading to Django 3.0 is the new async functionality getting in the way of `watch_transactions` blocking IO calls. I added code comments describing my thoughts

JakeUrban:
I've decided to table this upgrade until further notice. I don't want to make it any harder for anchors to upgrade to 1.0. We will however make `PolarisSameSiteMiddleware` optional for those anchors who don't support wallets that use iframes.

PR #287, opened 2020-08-20
Django 3.1 Upgrade

JakeUrban:
resolves stellar/django-polaris#236

Upgrades Polaris to install the latest `3.*` version

Also removes requirement of `PolarisSameSiteMiddleware`, which inserted cookie headers for clients using HTML iframes to allow session cookies to persist. With Django 3.1, Django supports setting the cookie header value through a Django API.

TODO:
- Update documentation
- Add cookie headers via Django API if we still want to support iframes. We may not want to do this since the SEP clearly recommends popups


JakeUrban:
I've decided to postpone the 3.0 upgrade, particularly because of the asynchronous limitations imposed on `watch_transactions`.

Issue #304, opened 2020-09-02
SEP-31: how to filter customer by SEP10 pubkey


yuriescl:
SEP-31 requires SEP-10 authentication and I assume that only a customer which belongs to the SEP-10 account can be returned in the response.
[CustomerIntegration.get()](https://github.com/stellar/django-polaris/blob/master/polaris/polaris/integrations/customers.py#L17) currently only receives the parameters passed to the `/customer` call, which might not contain the JWT with the SEP-10 pubkey. How can I use the SEP-10 account as a filter to avoid returning a customer that is not authenticated in the request?

JakeUrban:
This is an issue, you're right. I'll fix this bug before releasing 1.0

Issue #303, opened 2020-09-02
SEP-31 /customer: how to return a 404 response


yuriescl:
If `id` is passed to `/customer` and the customer does not exist, we should return a `404 Not Found` response.
[CustomerIntegration.get()](https://github.com/stellar/django-polaris/blob/master/polaris/polaris/integrations/customers.py#L17) currently only supports raising a ValueError, which will make Polaris return a `400` response when a `404` would be more appropriate.
I didn't find any mention about a `404` response in SEP-12 spec, is this a protocol limitation or should we return `400` in all cases?

JakeUrban:
`GET /customer` requests with an invalid `id` currently would return `400 Bad Request`. We could update this to add a 404 response, but it would only be used when `id`'s are specified. 

If they aren't, `GET /customer` would never return a 404 or 400, instead returning a 200 along with the fields required for the optional `type` passed.

JakeUrban:
If the `id` parameter was required, I would definitely support adding a 404 response. I'm not sure how I feel about adding one in this situation. What do you think @tomerweller ?

yuriescl:
> We could update this to add a 404 response, but it would only be used when `id`'s are specified

Yes this applies only when the `id` is passed.

yuriescl:
I'm also not allowing other filters to be passed along with `id`, since it's a unique value.

JakeUrban:
I think it may be the best solution to return `404`. Otherwise, clients won't be able to distinguish between ill-formed requests and non-existent users.

JakeUrban:
This too will also require Polaris 1.0 delays

PR #305, opened 2020-09-02
Added 404 response and always pass account to get()

JakeUrban:
resolves stellar/django-polaris#304
resolves stellar/django-polaris#303

### DIFF
--- a/polaris/polaris/sep12/customer.py
+++ b/polaris/polaris/sep12/customer.py
@@ -45,7 +45,9 @@ class CustomerAPIView(APIView):
             response_data = rci.get(
                 {
                     "id": request.GET.get("id"),
-                    "account": request.GET.get("account"),
+                    # anchors need to be able to filter out customers who weren't
+                    # created by the following SEP-10 JWT source account.
+                    "account": account,
                     "memo": request.GET.get("memo"),
                     "memo_type": request.GET.get("memo_type"),
                     "type": request.GET.get("type"),
@@ -54,6 +56,8 @@ class CustomerAPIView(APIView):
             )
         except ValueError as e:
             return render_error_response(str(e), status_code=400)
+        except ObjectDoesNotExist as e:
+            return render_error_response(str(e), status_code=404)
 
         try:
             validate_response_data(response_data)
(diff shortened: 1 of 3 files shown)


JakeUrban:
@yuriescl

polaris/polaris/sep12/customer.py
```diff
@@ -45,7 +45,9 @@ def get(account: str, request: Request) -> Response:
             response_data = rci.get(
                 {
                     "id": request.GET.get("id"),
-                    "account": request.GET.get("account"),
+                    # anchors need to be able to filter out customers who weren't
+                    # created by the following SEP-10 JWT source account.
+                    "account": account,
```
yuriescl, line 50: I think we need to pass two different accounts here, one would be the SEP-10 pubkey, and the other would be the `account` query parameter. They can be different, as mentioned [here](https://github.com/stellar/stellar-protocol/issues/676#issuecomment-656252392).

JakeUrban: They could be in the future but right now Polaris expects them to be the same.

JakeUrban: We could pass the JWT source account to the `CustomerIntegration.get()` function under the `authenticated_account` key, and leave `account` as `None` to differentiate between the request parameter and the JWT source account?

acharb (approved):
LGTM

polaris/polaris/sep12/customer.py
yuriescl: I still think we need two, it might be confusing to pass a value for the query parameter `account` and receive a response based on a different account (from SEP-10). I suggest either returning an error if they are different, or passing it to the integration function so the developer can decide to raise an error if they're different.

yuriescl: > We could pass the JWT source account to the CustomerIntegration.get() function under the authenticated_account key, and leave account as None to differentiate between the request parameter and the JWT source account?

Yes, sorry my previous reply was to your previous message.
I think it's better to pass both to the integration function.

JakeUrban: They will always match for now, but we can pass both so we can avoid the breaking change post-1.0

Issue #306, opened 2020-09-04
CustomerIntegration: delete should raise ObjectNotFound for 404


yuriescl:
[CustomerIntegration.delete()](https://github.com/stellar/django-polaris/blob/master/polaris/polaris/integrations/customers.py#L63) should raise `ObjectNotFound` instead of `ValueError` for 404 responses.
`ValueError` should be raised if the values are invalid.

JakeUrban:
`ValueError` will be replaced with `ObjectNotFound` in `.delete()`, since only 404 and 200 are returned for this endpoint.

PR #308, opened 2020-09-09
Change SEP-12 DELETE exception from ValueError to ObjectDoesNotExist

JakeUrban:
resolves stellar/django-polaris#306

Simple but breaking change. Updates the exception to match the other functions for SEP-12.


acharb (approved):
LGTM

PR #311, opened 2020-09-20
exit commands gracefully

yuriescl:
The management commands need to gracefully exit, otherwise they can be interrupted in the middle of transactions and the database might get inconsistent.
This pull request uses the native `signal` module to register the `SIGINT` and `SIGTERM` signals. When the process is either interrupted by Ctrl+C or killed, it finishes the current loop iteration and then exits.

Without a graceful exit, a common scenario that can cause issues is when the code is being deployed/updated on the server, and all the management commands get killed in order to be restarted to run the new code. This can cause the process to stop at a random point and some transactions might get into an inconsistent state in the database (went through but status wasn't updated, etc).

### DIFF
--- a/polaris/polaris/management/commands/check_trustlines.py
+++ b/polaris/polaris/management/commands/check_trustlines.py
@@ -18,6 +18,7 @@ class Command(BaseCommand):
     Create Stellar transaction for deposit transactions marked as pending trust, if a
     trustline has been created.
     """
+
     default_interval = 60
     _terminate = False
 
--- a/polaris/polaris/management/commands/check_trustlines.py
+++ b/polaris/polaris/management/commands/check_trustlines.py
@@ -1,3 +1,4 @@
+import signal
 import time
 
 from django.core.management.base import BaseCommand
@@ -17,31 +18,49 @@ class Command(BaseCommand):
     Create Stellar transaction for deposit transactions marked as pending trust, if a
     trustline has been created.
     """
+    default_interval = 60
+    terminate = False
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        signal.signal(signal.SIGINT, self.exit_gracefully)
+        signal.signal(signal.SIGTERM, self.exit_gracefully)
+
+    def exit_gracefully(self, sig, frame):
+        self.terminate = True
+
+    def sleep(self, seconds):
+        for i in range(0, seconds):
+            if self.terminate:
+                break
+            time.sleep(1)
 
     def add_arguments(self, parser):
         parser.add_argument(
             "--loop",
             action="store_true",
-            help="Continually restart command after a specified " "number of seconds.",
+            help="Continually restart command after a specified number of seconds.",
         )
         parser.add_argument(
             "--interval",
             "-i",
             type=int,
-            help="The number of seconds to wait before "
-            "restarting command. Defaults to 60.",
+            help="The number of seconds to wait before restarting command. "
+            "Defaults to {}.".format(self.default_interval),
         )
 
     def handle(self, *args, **options):
         if options.get("loop"):
             while True:
-                self.check_trustlines()
-                time.sleep(options.get("interval") or 60)
+                if self.terminate:
+                    break
+                self.check_trustlines(self)
+                self.sleep(options.get("interval") or self.default_interval)
         else:
-            self.check_trustlines()
+            self.check_trustlines(self)
 
     @staticmethod
-    def check_trustlines():
+    def check_trustlines(self=None):
         """
         Create Stellar transaction for deposit transactions marked as pending trust, if a
         trustline has been created.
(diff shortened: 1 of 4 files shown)


yuriescl:
93fd26a56280f3585096f37641eb4fa2a8fc9863 is a small bugfix in the default intervals

716352704ed844f03e528e42c6ba7411b2fc84b4 makes commands exit gracefully, otherwise the database can get inconsistent in case if SIGINT/SIGTERM signals

polaris/polaris/management/commands/check_trustlines.py
```diff
@@ -17,31 +18,49 @@ class Command(BaseCommand):
...
+                    break
+                self.check_trustlines(self)
+                self.sleep(options.get("interval") or self.default_interval)
         else:
-            self.check_trustlines()
+            self.check_trustlines(self)
 
     @staticmethod
-    def check_trustlines():
+    def check_trustlines(self=None):
```
yuriescl, line 63: @JakeUrban I'm not sure if passing an optional `self` to `@staticmethod` functions is a good practice, I did it like that to avoid too many changes in the code, specially in the tests.

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -59,32 +60,49 @@ class Command(BaseCommand):
...
+                    break
+                self.execute_deposits(self)
+                self.sleep(options.get("interval") or self.default_interval)
         else:
-            self.execute_deposits()
+            self.execute_deposits(self)
 
     @classmethod
-    def execute_deposits(cls):
+    def execute_deposits(cls, self=None):
```
yuriescl, line 105: @JakeUrban This doesn't look very optimal as well.

polaris/polaris/management/commands/check_trustlines.py
JakeUrban: Haha its not exactly good practice.

JakeUrban:
I definitely agree we need graceful exiting, but I need some time to understand the signal and terminate logic. Ideally we don't need to write `if self.terminate` or `if self` logic in a bunch of places.

polaris/polaris/management/commands/check_trustlines.py
yuriescl: Yeah. It was just a workaround for the mock tests, otherwise I'd remove the `@staticmethod` decorator from all the commands and receive `self` as a normal method parameter.

yuriescl: Maybe using something like `call_command` for calling the commands in the tests instead of `Command.check_trustlines()`, and remove the `@staticmethod`.

yuriescl:
> I definitely agree we need graceful exiting, but I need some time to understand the signal and terminate logic. Ideally we don't need to write `if self.terminate` or `if self` logic in a bunch of places.

I think we do need `if self.terminate` in specific places, since the idea of handling signals is to have determined places where the script can stop.
The `self` logic, as mentioned, was a (not clean) workaround just to avoid syntax errors in the tests, but it would make much more sense without the `@staticmethod` context, as `terminate` would be a internal state controller of the `Command` instance.

yuriescl:
Since the script sleeps, we need to check wether to terminate it or not without waiting for the sleep to finish, that's the reason for the custom `sleep` method, it checks every second.
`terminate` must be checked at the start of loop iterations, because the script cannot stop at a random point during the iteration.

JakeUrban:
@yuriescl after reading this over again I agree with your code. I'll update the tests and use it in 1.1.

JakeUrban:
@yuriescl if you could update the tests that would be greatly appreciated. I got a lot on my plate at the moment but the `self` issue shouldn't be too difficult to fix. I believe these functions need to be `static` so I can call them in tests, but as you said we could use `call_command`.

The problem with that approach is that it tests the whole command flow and would involve more mocking/setup. Ideally, `self.terminate` is communicated in some other fashion instead of per-object. Since there is only one object per process (for example, there's only one `Command` object per `watch_transactions` command call), I think we can use process-level variables.

This makes this very easy to do. We can define a `TERMINATE = False` module-level variable and update it to `True` when `exit_gracefully()` is called. We can then check `TERMINATE` and at the points you've already implemented and exit the program.

yuriescl:
@JakeUrban I replaced `self` with a terminate function, which returns `True` if the command must terminate.

yuriescl:
@JakeUrban I see you added a commit with module-level `TERMINATE`, I'm not sure how to proceed.

JakeUrban:
I think this is the way to go vs. passing terminate functions in our `Command` methods. Since we don't have to store the terminate boolean in the object, there's no reason not to keep the boolean at the module level.

JakeUrban:
I'm testing the commands now, are you alright with the implementation as it is right now?

yuriescl:
@JakeUrban I'm alright with module-level. The reason I didn't implement it is because I was thinking of `terminate` as an object state controller, not as a module global attribute.
Whatever you think is best on this case.

PR #312, opened 2020-09-20
SEP-12: pass sep10_client_account to put()

yuriescl:
`sep10_client_account` is being passed to `get()`, it should be passed to `put()` as well for consistency.

### DIFF
--- a/polaris/polaris/sep12/customer.py
+++ b/polaris/polaris/sep12/customer.py
@@ -95,6 +95,7 @@ class CustomerAPIView(APIView):
             customer_id = rci.put(
                 {
                     "id": request.data.get("id"),
+                    "sep10_client_account": account,
                     "account": account,
                     "memo": request.data.get("memo"),
                     "memo_type": request.data.get("memo_type"),


yuriescl:
@JakeUrban If this gets accepted, the test will need to be updated, it's currently failing due to the new field.

JakeUrban (commented):
Looking back on the decision to add `sep10_client_account` before the possibility of a difference in key values, I probably should have waited to add the attribute. The change wouldn't have been breaking, it would have been additive.

At this point I'm not against adding this attribute since its already included in the `.get()` call but its not helpful for the developer yet.

yuriescl:
@JakeUrban As far as I remember, the reason for adding `sep10_client_account` it could differ from `account` (in the future?).
Can you elaborate on why you're enforcing them to be the same?

JakeUrban:
Because if the account values weren't the same today it would be an invalid request. Yes, they could be different in the future, but as I said in my previous comment I should have waited until the difference was possible.

yuriescl:
> Because if the account values weren't the same today it would be an invalid request.
What would make it an invalid request?

JakeUrban:
If an account X was authorized via SEP-10 and tried to register a customer with another service under account Y, the service _must_ reject that request because X doesn't have permission it create customers in the name of Y. 

`M-addresseses` are the only exception to this rule, but they do not exist in the Stellar ecosystem today.

yuriescl:
> `M-addresses` are the only exception to this rule, but they do not exist in the Stellar ecosystem today.

Oohh, ok I thought they were already active. Makes total sense to invalidate the request then. Thanks for the info!

JakeUrban:
I'm going to hold off on this for now until that happens.

PR #313, opened 2020-09-21
fixed interactive withdraw url

yuriescl:


yuriescl:
@JakeUrban please add this to v1.1 if possible

Issue #289, opened 2020-08-21
Support Multi-sig Assets


yuriescl:
`Asset` holds a `distribution_seed` which is used to derive the pubkey, and the pubkey is used by responses and for watching transactions.
This makes it difficult for the Anchor if it wants to have a separate account just for receiving the payments (this would not require Polaris to know the private key).
Some Anchors have multiple distribution accounts and might not want to store all private keys in the server for security purposes.

Two questions:
- Is it currently possible to have multiple distribution accounts setup in Polaris?
- Is it currently possible to setup multi-sig distribution accounts in Polaris, in a way that it would allow an external service to add signatures to a transaction and allow it to proceed?

Both scenarios described above might be possible but I'm not sure if it would be worth to workaround the current Polaris instead of having it natively supported by Polaris.

JakeUrban:
Heres how I'm thinking about a potential solution:

A polling-approach where clients go through a SEP-10ish dance to authenticate with and poll a custom Polaris SEP-24 endpoint could work. If proven to be useful and effective, this dance could be added to the SEP-24/SEP-10 spec. Specifically, this flow would benefit from a `pending_signatures` external SEP-24 status.

Here is what that flow might look like without the `pending_signatures` status:

1. A deposit is been received off-chain and the internal transaction record is found
2. Polaris will check if the transaction requires more than one signature based on the `signers` object on the distribution account of transaction's asset
    - These objects will likely be pulled from Horizon on application startup
3. If additional signatures are required, the transaction's external status is set to `pending_external`, a new Polaris `Transaction.pending_signatures` column is set to `True`, and a serialized `TransactionEnvelope` is saved to a new Polaris' `Transaction.envelope` column
4. Client applications in possession of a secret key for one of the distribution account's signers can authenticate with the Polaris service by going through a modified SEP-10 flow using the `GET /signers/auth?account=<account>` endpoint to receive a challenge transaction for the distribution `account`.
5. If the signature on the challenge sent back to the anchor via `POST /signers/auth` endpoint is from a signer of the distribution account, a SEP-10 JWT token should be returned by the anchor
6. The client can then use the SEP-10 JWT token to make requests to a `GET /signers/transactions` endpoint to get the `Transaction.envelope` XDRs that require the client's signature. 
7. These envelopes can be signed and `POSTed` back to the `/signers/transactions` endpoint. 
8. Once a transaction has enough signatures Polaris will change the status to `pending_anchor` and submit the transaction on the Stellar network

[...]

PR #314, opened 2020-09-21
Multi-signature assets

JakeUrban:
resolves stellar/django-polaris#289

This PR refactors `create_stellar_deposit()` into the following functions:
- `create_stellar_deposit()`
    - `get_or_create_stellar_account()`
    - `create_transaction_envelope()`
    - `submit_stellar_deposit()`

It also allows anchors to place transactions in a `pending_signatures` state where their external status is `pending_anchor` while internally the anchor collects signatures and writes the envelope to `Transaction.envelope`. 

It is up to the anchor how the envelope written to the DB is signed, but there are really two categories of solutions:
- _Pull_ models that require signing parties to request the transactions that are pending signatures
    - the signing parties would detect, sign, and submit transaction envelopes back to the anchor service
- _Push_ models that require the anchor to send the envelope to signing parties
    - the signing parties would receive, sign, and return the transaction envelope to the anchor service

When `Transaction.envelope` has enough signatures to submit to the Stellar network, the anchor should update `Transaction.pending_signatures` to `False` and `Transaction.status` to `Transaction.STATUS.pending_anchor`. Polaris will add these transactions to the list of `Transaction` objects ready to be submitted on the Stellar network.

These multi-signature transactions are submitted with channel accounts as the top-level source so envelopes can be serialized, signed, and submitted independently of distribution account's sequence number. The anchor should provide existing channel accounts via the `DepositIntegration.channel_keypair_for_multisig_transaction()` integration function. These accounts will be saved to the `Transaction.channel_secret` and `Transaction.channel_account` columns.

The anchor is responsible for cleaning up channel accounts, and should likely be created per-call to the integration function. A good spot to clean up channel accounts is in `DepositIntegration.after_deposit()`.


JakeUrban:
@yuriescl

yuriescl (approved):
👍

JakeUrban:
@yuriescl before this is released in 1.1 can you test this? I did my testing and got it working but want to double check.

yuriescl:
@JakeUrban 
Getting an error when running `poll_pending_deposits`:
```
Traceback (most recent call last):
  File "manage.py", line 21, in <module>
    main()
  File "manage.py", line 17, in main
    execute_from_command_line(sys.argv)
  File "/home/yuri/.virtualenvs/polaris/lib/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/home/yuri/.virtualenvs/polaris/lib/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/yuri/.virtualenvs/polaris/lib/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/yuri/.virtualenvs/polaris/lib/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/home/yuri/.virtualenvs/polaris/lib/python3.8/site-packages/polaris/management/commands/poll_pending_deposits.py", line 115, in handle
    self.execute_deposits()
  File "/home/yuri/.virtualenvs/polaris/lib/python3.8/site-packages/polaris/management/commands/poll_pending_deposits.py", line 178, in execute_deposits
    transaction.envelope = create_transaction_envelope(
  File "/home/yuri/.virtualenvs/polaris/lib/python3.8/site-packages/polaris/utils.py", line 282, in create_transaction_envelope
    transaction.amount_in - transaction.amount_fee,
TypeError: unsupported operand type(s) for -: 'decimal.Decimal' and 'NoneType'

```
This is caused by [create_transaction_envelope](https://github.com/JakeUrban/django-polaris/blob/multisignature-assets/polaris/polaris/utils.py#L280) **not** calculating `amount_fee` automatically in the same way as [execute_deposit](https://github.com/JakeUrban/django-polaris/blob/multisignature-assets/polaris/polaris/management/commands/poll_pending_deposits.py#L46), so `amount_fee` can end up being `None` in that `round` call.

yuriescl:
@JakeUrban This error is also happening:
```
File "/home/yuri/.virtualenvs/polaris/lib/python3.8/site-packages/polaris/utils.py", line 282, in create_transaction_envelope
    transaction.amount_in - transaction.amount_fee,
TypeError: unsupported operand type(s) for -: 'decimal.Decimal' and 'float'
```

Which happens when `transaction.amount_fee` is set in `process_sep6_deposit`, for example:
```
transaction.amount_fee = 1.5
```
Since the assigned value has type `float` and `transaction` is passed directly between the functions, it isn't converted to `Decimal`.
A possible fix is to convert both amounts to either `Decimal` or `float` when calling [round](https://github.com/JakeUrban/django-polaris/blob/multisignature-assets/polaris/polaris/utils.py#L281).

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -135,9 +141,45 @@ def execute_deposits(cls):
...
+                transaction.status = Transaction.STATUS.pending_anchor
+                asset_code = transaction.asset.code
+                if asset_code not in distribution_accounts:
+                    distribution_accounts[
+                        asset_code
+                    ] = settings.HORIZON_SERVER.load_account(
+                        transaction.asset.distribution_account
+                    )
+                transaction.envelope = create_transaction_envelope(
+                    transaction, distribution_accounts[asset_code]
```
yuriescl, line 179: @JakeUrban why not pass `asset.distribution_account`? why is there a need for this `distribution_accounts` dict?

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -135,9 +141,45 @@ def execute_deposits(cls):
...
+        # TODO: PR REVIEWERS, should we ask the anchor which transactions should
+        #  be added to the list of ready transactions or try to query them ourselves
+        #  like so?
+        #  We should definitely provide an integration function or explict
+        #  Transaction boolean column if the query below could in theory collect
+        #  transactions that are not ready. From my understanding, all Transaction
+        #  objects collected with the attribute values below _would_ be ready.
+        #  The idea is that anchors change the _status_ and _pending_signatures_
+        #  field to signal to Polaris that the object is ready.
+        ready_transactions.extend(
```
yuriescl, line 156: @JakeUrban I can see two scenarios:
1. Polaris does this query and validates the envelope signatures (checks for duplicate signatures and sums up the weights). On this case, there's no integration function necessary and the Anchor must set `pending_signatures` to `True` to signal that the transaction is ready.
2. Polaris calls an integration function to get the ready transactions. On this case, Polaris doesn't check anything and just submits them. The Anchor would be responsible for all checks.

I don't see why Polaris would have an integration function and do the checks as well, it would add restriction on top of the integration.

yuriescl:
@JakeUrban It would be useful to have documentation regarding the `pending_signatures` and `envelope` fields in the following places:
[DepositIntegration.process_sep6_request](https://django-polaris.readthedocs.io/en/stable/sep6_and_sep24/index.html#polaris.integrations.DepositIntegration.process_sep6_request) - Anchor can set the initial `pending_signatures` here.

[DepositIntegration.after_form_validation](https://django-polaris.readthedocs.io/en/stable/sep6_and_sep24/index.html#polaris.integrations.DepositIntegration.after_form_validation) - Anchor can set the initial `pending_signatures` here.

[RailsIntegration.poll_pending_deposits](https://django-polaris.readthedocs.io/en/stable/sep6_and_sep24/index.html#polaris.integrations.RailsIntegration.poll_pending_deposits) - Add multi-sig instructions

In the documentation it might be worth mentioning Django REST Framework in case the Anchor wants to build an API to manage the transactions, specially the multi-sig.
There are serializer classes built-in in Polaris and they can be used to build APIs quickly using DRF. Plus the fact that DRF is already a dependency of Polaris.

```
from rest_framework import serializers, routers
from polaris.models import Transaction

class TransactionSerializer(serializers.ModelSerializer):
    class Meta:
        model = Transaction
        fields = '__all__'

class TransactionViewSet(viewsets.ModelViewSet):
    serializer_class = TransactionSerializer
    queryset = Transaction.objects.all()
    authentication_classes = []

router = routers.SimpleRouter()
router.register('transactions', TransactionViewSet, basename='transactions')

urlpatterns = router.urls
```

polaris/polaris/management/commands/poll_pending_deposits.py
JakeUrban: I agree, we can handle it.

polaris/polaris/management/commands/poll_pending_deposits.py
JakeUrban: `asset.distribution_account` is a public key string. `settings.HORIZON_SERVER.load_account()` returns an `stellar_sdk.account.Account` object, which is whats required by `create_transaction_envelope()`.

Currently we're caching previously requested `Account` objects for the respective `distribution_account`s. But I'm realizing we need to refresh the accounts to get updated sequence numbers.

JakeUrban: Actually, the sequence number will likely be the same if the transaction envelope created in the last iteration has not been submitted. 

This is an issue because Polaris never attempts to maintain an internal sequence number for an account, it always just pulls it from Horizon. But since we're not submitting these transactions right away, the sequence numbers will be wrong by the time they are submitted.

This means Polaris needs to maintain an internal sequence number for every anchored asset's distribution account.

JakeUrban: This also means that ideally, Polaris is the only service that submits transactions from the asset's distribution accounts. If this isn't the case, Polaris will have to request the correct sequence number from Horizon and update existing envelopes with sequence numbers _less than the correct sequence number_.

JakeUrban: Turns out updating sequence numbers would invalidate signatures, which makes sense. The solution is to submit the transaction from a Stellar account _derived_ from the distribution account. 

If every multisig transaction has uses a derived source account, we can:
1. Fund the account
2. Make the payment using that account
3. Merge the account back to the distribution account

This ensures no mulisig transaction will be rejected for sequence number-related reasons. The downside is added performance cost when submitting these transactions.

acharb (approved):
LGTM

polaris/polaris/management/commands/poll_pending_deposits.py
yuriescl: @JakeUrban If I understood correctly, you're suggesting that for each transaction:
1. create a new account derived from `distribution_account`
2. use the new account as the channel source and initially sign the envelope with it
3. let the anchor add the signatures to the envelope
4. submit the transaction
5. merge the new accoun back to the distribution account

That works, since it removes the need to worry about the sequence numbers and also allows multiple Polaris instances using the same distribution account.
Channels seem very useful, maybe it would be interesting to have channel support in Polaris in the future for batch transaction submitting.

JakeUrban:
@yuriescl I'm working on implementing the channel account functionality and I'm realizing that if an asset's distribution account requires multiple signatures for payment transactions, the anchor must collect signatures _for the transaction that creates the channel account._ 

However this is not feasible because of the exact same issue that makes the channel accounts necessary: The asset's distribution account sequence number will be greater than the sequence numbers on these transaction envelopes by the time they're submitted.

So, the strategy of having Polaris create a channel account and merging it back after the deposit is not possible, realistically. Instead, I suggest that the anchor provides a pre-existing account that can be used as the top-level source account for deposit transactions. The sole purpose of this account should be to pay for Polaris' deposit transaction fees and to allow Polaris to consume its sequence numbers.

The anchor would be able to provide an account per-deposit-transaction and should control how long it exists. Anchors would be free to use a persistent non-distribution account that holds an appropriate XLM balance to pay fees _or_ create an account per deposit transaction and funding it appropriately. What do you think of this approach?

yuriescl:
> The anchor would be able to provide an account per-deposit-transaction and should control how long it exists. Anchors would be free to use a persistent non-distribution account that holds an appropriate XLM balance to pay fees or create an account per deposit transaction and fund it appropriately. What do you think of this approach?

@JakeUrban Works for me

JakeUrban:
I also want to note an assumption I'm trying to get rid of in Polaris: every asset's distribution account has a non-zero master signer weight.

If the weight of the distribution account's master signer is zero, Polaris shouldn't sign the transaction with the distribution account seed before serializing the envelope or validating it before submitting. This is not something Polaris does yet, and is something I'd like to support with this functionality.

Will your distribution account(s) have non-zero master weight signers?

JakeUrban:
@yuriescl This is largely done, I just need to polish and add documentation. Take one last look and I'll write that up and merge.

yuriescl:
> Will your distribution account(s) have non-zero master weight signers?

@JakeUrban Maybe.

JakeUrban:
I've made it slightly easier to use, I'll explain how it works tomorrow.

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -138,17 +125,96 @@ def execute_deposits(cls):
...
+                )
+            thresholds = json.loads(transaction.asset.distribution_account_thresholds)
+            if not (
+                master_signer and master_signer["weight"] >= thresholds["med_threshold"]
+            ):
+                # master account is not sufficient for
+                transaction.pending_signatures = True
+                transaction.status = Transaction.STATUS.pending_anchor
+                transaction.save()
+                if transaction.channel_account:
```
yuriescl, line 169: @JakeUrban Maybe the channel account should be obtained only from calling `rdi.channel_keypair_for_multisig_transaction`. It can be confusing to be able to set `transaction.channel_account` and also have an integration function to provide it.
Polaris should be the one setting `transaction.channel_seed` with the value obtained from the integration function.
This also makes it easier for the Anchor to handle channel accounts, since the entire logic can reside inside the integration function, no need to worry about storing it in the transaction beforehand.

polaris/polaris/models.py
```diff
@@ -86,11 +117,17 @@ def get_db_prep_value(self, value, *args, **kwargs):
...
 class Asset(TimeStampedModel):
-    """
-    .. _Info: https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0024.md#info
-
-    This defines an Asset, as described in the SEP-24 Info_ endpoint.
-    """
+    def __init__(self, *args, **kwargs):
+        """
+        Does the usual __init__() actions and ensures the asset's distribution account
+        information is update to date if it hasn't been pulled from Horizon since
```
yuriescl, line 123: typo here, should be "up to date"

polaris/polaris/management/commands/poll_pending_deposits.py
yuriescl: Or maybe this is just to make it easier for writing tests, since you can set `channel_seed` directly. Nevermind the comment.

JakeUrban: I actually agree. `rdi.channel_keypair_for_multisig_transaction` will not be passed a `transaction` object now. Instead it should return a new/fresh channel account every time it's called. Maybe it should just be called `rdi.create_channel_account()`?

JakeUrban: Not that there are only two times a channel account would be used per-transaction:
- When the deposit transaction is submitted (obviously)
- But also before that, if the destination account doesn't exist, a channel account will be used as the source account of the TX while the asset's distribution account is the source of the payment operation.

Since this 2nd (1st if going in-order of the flow) channel account isn't actually for the deposit transaction, `rdi.after_deposit()` is never called for that channel account. So I think we need a clean-up function to merge outstanding channel accounts.

`rdi.merge_channel_account()`? Called after Polaris submits the TX for which the channel account was created for.

JakeUrban: Then `rdi.after_deposit()` wouldn't be the place you merge channel accounts anymore.

JakeUrban: Or actually, maybe we continue with one channel account per transaction, not two. If we choose to go that route, then `transaction.channel_account` makes sense to have because it will always be used for transactions related to the actual deposit transaction the object represents.

We'd still have `create_channel_account()` and `merge_channel_account()`, but they would be called when the first TX needs to be submitted on behalf of the deposit and when the deposit has completed or failed, respectively. 

If the deposit failed, it should be the anchor if they want to merge the channel account back or wait until the deposit is retried after resolving an issue. When a previously failed deposit is retried after manual intervention, `merge_channel_account()` would be called again.

JakeUrban: Would it make more sense for `create_channel_account()` not to return anything and instead be passed the `transaction` object, so the anchor could assign `transaction.channel_seed`? Polaris would still validate the existence of the account.

JakeUrban: `merge_channel_account` lines up exactly with `after_deposit()` so we're not going to do `merge_channel_account()` and instead there will be added emphasis on `after_deposit()`.

yuriescl: @JakeUrban `rdi.create_channel_account()` would require on-demand multi-sig, which is what we're trying to avoid.
Since the need for channel accounts on this multi-sig feature is due to the idea of allowing the Anchor do polling instead of on-demand, I think we should drop the polling support and do on-demand multi-sig only, in which the Anchor **has** to sign the envelope when Polaris tries to submit it. Otherwise it's a circular logic and sequence numbers will always be an issue.

JakeUrban: I agree it would be simpler to require a push method but I think thats pretty restrictive.

I'm not sure what you mean by "on-demand multisig". As described in some comments I have locally (changes I haven't pushed):
```
    If the transacted asset's distribution account does require non-master signatures, the anchor
    should save a keypair of a pre-existing Stellar account to use as the channel account via
    DepositIntegration.create_channel_account().

    This channel account must only be used as the source account for transactions related to the
    ``Transaction`` object passed. It also must not be used to submit transactions by any service
    other than Polaris. If it is, the outstanding transactions will be invalidated due to bad
    sequence numbers. Finally, the channel accounts must have a master signer with a weight greater
    than or equal to the medium threshold for the account.
```

JakeUrban: This last line: `Finally, the channel accounts must have a master signer with a weight greater than or equal to the medium threshold for the account` is important. 

It means the asset's distribution account requires multisig, but the channel accounts intentionally don't. Channel accounts should have master account that can sign for its medium-threshold transactions. This is totally fine because the actual deposit transaction still requires the distribution account's signatures. Meaning, its not any less secure.

The transaction that creates the destination account, if it doesn't exist, is the one that doesn't require the distribution account's signatures, because the channel account is paying for the created account's XLM reserve. This is also not a security concern because the channel account is created on-the-fly per-transaction. It should not have a significant balance of _any_ asset and should be destroyed as soon as the deposit has been facilitated.

Does this address your issues with channel accounts?

yuriescl: @JakeUrban 

> I agree it would be simpler to require a push method but I think thats pretty restrictive.

Me too.

> I'm not sure what you mean by "on-demand multisig"

On-demand means that Polaris needs the return value right now. For example, if the multi-sig for deposit was on-demand, the Anchor would have to add all signatures in a single function call ("push" method as you described previously) so Polaris could submit the envelope to horizon instantly.

> the anchor should save a keypair of a pre-existing Stellar account to use as the channel account

If the keypair is per-transaction, then I imagine the Anchor would have to create it somewhere like `process_sep6_deposit()`, and on this case it would have to add all signatures _on-demand_ (as described above) and store the keypair somewhere attached to this transaction, likely in a related model. That's why I said it would require on-demand multi-sig, and if the Anchor has, at some point, to do on-demand multi-sig, then I don't see why it wouldn't do on-demand for all cases, inclusing sending the payment to the user.

However, I do see a scenario where there's no need for on-demand multi-sig. The Anchor can simply not allow Polaris to process the transaction until it creates the channel account. This means the Anchor doesn't return the transaction in `poll_pending_deposits()` until there's a channel account ready for that transaction.

JakeUrban: > If the keypair is per-transaction, then I imagine the Anchor would have to create it somewhere like process_sep6_deposit(), and on this case it would have to add all signatures on-demand (as described above) and store the keypair somewhere attached to this transaction, likely in a related model. 

All anchors, regardless of if they're implementing SEP-6 or SEP-24, will implement `create_channel_account(transaction)`, which should create the channel account on Stellar. It should be simple, non-multisig account. 

Polaris will call this function when one of two things happen, assuming the user sent the off-chain deposit to the anchor and the anchor has detected/matched that deposit with the internal record:
 - If the client's Stellar account doesn't exist, `create_channel_account()` is called to create it. The channel account is the one that pays the XLM reserve for the account created. This channel account must be a simple master-signer account, AKA no multisig.
- If the clients Stellar account does exist, then `create_channel account()` will be called for `transaction` so it can be used to make the actual deposit transaction to the client's Stellar account

So a channel account will be used for 1-2 transactions, depending on if the Stellar account exists. If the asset's distribution account is not configured for multisg, channel accounts will not be used.

Since the channel account is per-transaction, the channel account will be stored on `Transaction.channel_seed` and `Transaction.channel_account`.

JakeUrban: Let me write up the documentation and implement it in the anchor server so you can see how it would work for SEP-24 and SEP-6. I think we have what we need now.

JakeUrban: I think the bottom line in response to your concern is that all channel accounts _must_ be created on-demand and merged back to whatever account is creating them when the TX has been successfully submitted on the network.

yuriescl: @JakeUrban 

> All anchors, regardless of if they're implementing SEP-6 or SEP-24, will implement create_channel_account(transaction), which should create the channel account on Stellar. It should be simple, non-multisig account.

Yes, but to create the channel account, multiple signers are required. At least that's what I understood from this statement:

> I'm working on implementing the channel account functionality and I'm realizing that if an asset's distribution account requires multiple signatures for payment transactions, the anchor must collect signatures for the transaction that creates the channel account.

So, in `create_channel_account()`, the Anchor has to instantly collect all signatures otherwise it won't be able to create the channel account. Is that correct?

yuriescl: I might be a bit lost on the restrictions we have with multi-sig. This is my current understanding:
1) to avoid issues with sequence numbers, we need channel accounts
2) the asset distribution account must create the channel account for each transaction, because the channel account must be derived from the distribution account
3) if the distribution account required multi-sig for its transaction, then it needs to collect signatures in order to **create** the channel account
4) once the channel account, is created, it will be used to send payments to the user and will be the source account of all the payments
  * when the channel account creates the envelope to send the payment to the user, does is require multi-sig on that envelope?

I'm not sure exactly where multiple signatures are required in all this process, that's why my ideas might seem a bit over complicated.

JakeUrban: Close, but yes you are overcomplicating it. I also didn't help by making comments as I think through the process myself.

> Yes, but to create the channel account, multiple signers are required.
> 2. the asset distribution account must create the channel account for each transaction, because the channel account must be derived from the distribution account

These are both false, the distribution account does not need to be derived. It can be an independent account that is not created from the asset's distribution account. This gets around the need for multiple signers in order to create the channel account. 

Originally, I thought it would make sense to use derived accounts and use the distribution account to create the destination account, but when I realized the issue of still needing signatures, I decided to remove that constraint.

JakeUrban: > when the channel account creates the envelope to send the payment to the user, does is require multi-sig on that envelope?

It will need multisig if the distribution account is used to send the payment.

yuriescl: @JakeUrban Thanks for the clarification. Please correct me if I'm wrong:
1. Polaris needs a source account to send payments
2. Anchor wants to require multiple signatures to send payments
3. To allow multiple signers on a transaction, Polaris needs to store the envelope and let the Anchor sign it whenever it wants
4. (3) Introduces the issue with sequence numbers
5. To work around the issue with sequence numbers, Polaris allows the Anchor to use an individual stellar account for each transaction, which is called channel account
6. The Anchor, for each transaction, would create and fund a channel account (which is temporary and will be merged back after the payment is sent?)
7. Polaris, when creating the deposit envelope, checks if the asset distribution account requires multiple signers. If yes, it sets `pending_signatures = True` and relies on the Anchor to provide a channel account for the transaction. The Anchor must provide a channel account instantly (see 7.3 below) when `rdi.create_channel_account()` is called.
- 7.1. Is it possible for the Anchor to create a channel account that doesn't require multiple signers to send payments?
- 7.2. If (7.1) is true, then I imagine that, the envelope that creates the channel account needs to be signed by the signers required initially on (2), correct?
- 7.3. Since the channel account must be instantly created and returned in `rdi.create_channel_account()`, and if (7.2) is true, then inside `rdi.create_channel_account()` the Anchor must implement some mechanism to do multi-signatures by requesting them using the push model, correct?
  * Push models that require the anchor to send the envelope to signing parties, the signing parties would receive, sign, and return the transaction envelope to the anchor service
- 7.4. If (7.3) is true, then Polaris would be requiring instant ("on-demand") multi-sig on `rdi.create_channel_account()` but then allowing "not instant" multi-sig on `transaction.envelope`. That's the scenario I was referring to in here:
  * If the keypair is per-transaction, then I imagine the Anchor would have to create it somewhere like process_sep6_deposit(), and on this case it would have to add all signatures on-demand (as described above) and store the keypair somewhere attached to this transaction, likely in a related model. That's why I said it would require on-demand multi-sig, and if the Anchor has, at some point, to do on-demand multi-sig, then I don't see why it wouldn't do on-demand for all cases, including sending the payment to the user.

JakeUrban: 1 through 7 are completely accurate (yes to 6).

> 7.1. Is it possible for the Anchor to create a channel account that doesn't require multiple signers to send payments?

It is a _requirement_ that channel accounts have a master signer whose weight is greater than it's medium threshold. That means that if the channel account wants to send XLM (or any asset) _from its own account balance_, it can do so by signing the transaction with the single account public key secret.

I know this is probably confusing because we're using channel accounts specifically for multisig transactions. So when do transactions using channel accounts need multiple signatures?

Every Stellar transaction has a source account. In this context, this is the channel account we're talking about, _which only has the master signer on the account_. 

**Optionally, many (all?) Stellar operations within a transaction can have their own source account that differs from the transaction-level source account.**

This is what we would do for deposit (payment) transactions in SEP-6 and SEP-24 when the asset's distribution account requires multiple signatures. 

Let's say asset MULT requires 2 signatures: one of which is the master signature and the other is an extra signer required to reach the medium threshold. This deposit transaction envelope must have the following structure before submitting:

TransactionEnvelope

- source: `channel account public key`
- operations:
    - payment:
        - source: `asset distribution account public key`
        - asset_code: `MULT`
        - asset_issuer: `asset issuer account`
- signatures:
    - `channel account master key signature`
    - `asset distribution account master key signature`
    - `asset distribution account extra signer key signature`

And remember, creating the channel account for a multisig transaction should _not_ be done by the distribution account, because the distribution account requires multiple signatures. Channel accounts must be created as a result of the `rdi.create_channel_account(tranaction)` function call, so use an account with a single signer to fund channel accounts.

I would set up a system where funds are sent from the distribution account to a _funding_ account periodically, say every hour. This funding account should have a single master signer so it can fund channel accounts for Polaris on-demand. It could also be the account funds are merged back to after the deposit completed (or funds could be merged directly back to the distribution account).

There are other ways you could setup accounts so that channel accounts can be created on-demand, but the idea is you need channel accounts to be created on demand from `create_channel_account()` and presumably this must be done with a non-multisig funding account, AKA not the distribution account.

Let me know if you have any questions about the above ^. I'm going to talk now about how channel accounts are _sometimes_ used twice per-deposit transaction.

**There's a special case: when the destination account doesn't exist**

If the destination account for the Stellar deposit doesn't exist, the anchor should create (sponsor) the account for them. This is typically done with the asset's distribution account funds. But when distribution accounts require multisig, the channel account for the transaction can be created early and used to fund the user's destination account.

So when the destination account for the deposit doesn't exist, a transaction like so is submitted:

TransactionEnvelope

- source: `channel account public key`
- operations:
    - create_account:
        - source: `channel account public key`
        - starting_balance: `minimum XLM starting balance` (2.01?)
- signatures:
    - `channel account master key signature`

Note that the transaction doesn't need multiple signatures. Thats because this transaction doesn't involve an account that requires multiple signatures.

So the channel account is used for this purpose and is kept around for when it needs to be used for the actual payment transaction to the destination account. Again, once payments are made, the channel account is merged back using `after_deposit()`. It should be merged back to the distribution account or the account that funded the channel account.

Hopefully I've addressed your questions and outlined how this can work. I'm curious to hear your thoughts. Obviously, if you want to implement a push model (immediate signing my additional signers), you can collect signatures quickly by calling an internal non-Polaris function and changing the `transaction.pending_signatures` back to `False` after Polaris makes `transaction.pending_signatures = True`.

yuriescl: @JakeUrban I think I got the idea. The key information was about the transaction envelope source, which can differ from the source of each operation in the envelope.
So the funding account just needs to have a minimal XLM balance (doesn't even need other balances) for creating channel accounts on-demand.
The payment envelope signatures can be added via polling by the Anchor, as previously planned (when channel accounts weren't involved).
I really appreciate the patience and time to write up the explanation, thanks a lot! :)

JakeUrban: No worries! I want to make sure I get this use case right. One thing you mentioned just now:

> The payment envelope signatures can be added via polling by the Anchor, as previously planned (when channel accounts weren't involved).

I think you're referring to the poll model the anchor could implement so clients can detect and add signatures. But you don't necessarily have to do that. You can have a push model where you have a separate `notify_signers` process that checks for transactions in `pending_signature=True` and makes API calls or does whatever it needs to do to get the signatures immediately before returning.

Its up to you what the process is for collecting signatures, Polaris just creates the initial envelope and waits for the transaction to not be `pending_signatures=True` anymore.

yuriescl:
@JakeUrban There was a function called `additional_signatures_needed`, which returns `True` if the envelope requires more signatures. It would be useful to restore it so Anchors can use it to check when to set `pending_signatures = True`. I'd say it's not that easy to figure out how to implement that check, and I imagine Anchors would appreciate that utility function.

JakeUrban:
Sure, I can add it back. The check Polaris cares about is whether or not it can use the account's master signer or if that won't be enough. This is done like so:
```
master_signer = None
if transaction.asset.distribution_account_master_signer:
    master_signer = json.loads(transaction.asset.distribution_account_master_signer)
thresholds = json.loads(transaction.asset.distribution_account_thresholds)
if not (master_signer and master_signer["weight"] >= thresholds["med_threshold"]):
```
Note that the distribution account could not have a master signer, which would be the case of the weight was 0.

The `additional_signatures_needed()` function would be a bit trickier since I'd have to figure out which threshold value the signatures need to meet and the weight of each signature. For example, if we're doing a multisig payment transaction with a channel account and distribution account as sources on the TX and Op respectively, I don't need to check the channel account's threshold, I need to check the distribution account's threshold, since thats the account making the payment.

yuriescl:
Oh it's more complicated than I thought. I guess that's one more reason to have it, specially for reference purposes.

JakeUrban:
I feel like it may make sense to let the developer define the signature weight detection solution. There are a lot of possible transaction configurations and I don't want Polaris taking responsibility for "approving" transactions to be submitted.

JakeUrban:
I think this is ready to be merged @yuriescl , do you want to check again or can I do it?

yuriescl (approved):

yuriescl:
> I think this is ready to be merged @yuriescl , do you want to check again or can I do it?

@JakeUrban Code looks good. Once this is released I'll test multi-sig on testnet and let you know if there are any issues.

Issue #299, opened 2020-08-26
Upgrade JWT used for interactive session to OTT


JakeUrban:
Polaris uses a separate & short-lived JWT to authenticate users when opening interactive URLs provided by the anchor. However, this JWT could be used to open many interactive windows with the same URL as long as the JWT expiration had not passed.

The goal of this change is to limit number of times the JWT can be used without being tied to a web session to once. 

The first time the interactive URL is opened, Polaris does not have yet have a session cookie it can map to the passed JWT. After Polaris creates and returns one in the webpage response, the JWT will no longer work. From thereon out, the session cookie is the only credential that can re-open an in-progress interactive flow.

JakeUrban:
I just realized this change limits the receiver of the URL to only open the window once. If closed, a new transaction must be initiated. This is already the case when the current short-lived JWT expires and the window containing the session cookie is closed.

yuriescl:
I agree with this change, but Wallets should be aware that they won't be able to reopen the same interactive URL twice.

About sessions, they can [break on iframes](https://github.com/stellar/stellar-protocol/issues/475) due to browser security restrictions on cookies.
I'm not sure if sessions are needed for Polaris. A short-lived JWT could be used instead.

As a side note (not directly related to this issue), embedded browser instances like [WKWebView](https://developer.apple.com/documentation/webkit/wkwebview/) can have issues with KYC processes. iOS for example doesn't allow certain camera operations inside web views, and I've seen this break real-time selfies which are required by some Anchors due to legislation. This forced the Wallet to open Safari with the interactive URL, so this scenario (Wallet opens browser with interactive URL) should also be taken into consideration.

JakeUrban:
We modified SEP-24 to suggest popups and removed the mention of iframes. Popups persist session cookies in its own session instead of piggybacking off the parent of the iframe's context.

While session cookies are not explicitly supported or encouraged in the SEP, it is a very easy and widely supported way of mapping session data to users interacting the site through a browser. 

JWT's on the other hand should only be used as a session token when it can be included in request headers or encrypted post bodies. If they're allowed in URLs, they _must_ be short-lived or OTT since the URL can easily be sent or sniffed by other parties who may go through a KYC and transaction flow with the original users' Stellar account.

Thats why Polaris uses a short-lived JWT; _to get a persistent `sessionid` response header_ in the popup's context. Once the popup has the session cookie, the JWT is no longer needed and but is currently still usable within the JWT's expiration.

[...]

PR #315, opened 2020-09-22
Disallow including the token after previously being authenticated

JakeUrban:
resolves stellar/django-polaris#299

This makes the interactive JWT used a one-time token. If included in the URL of an interactive webapp GET request after being authenticated in a prior GET request, the server should reject the token. This ensures only one window can be opened.


acharb (approved):
LGTM

Issue #319, opened 2020-09-28
Deprecation: JS script integrations


JakeUrban:
See https://github.com/stellar/django-polaris/issues/318, which will be introduced in 1.2 but will become the only option when 2.0 is released. JS Script integrations will be removed in 2.0 and deprecated in 1.2 onward.

Issue #318, opened 2020-09-28
Anchor-defined head and body templates


JakeUrban:
- Modify `base.html` by adding `extra_head` and `extra_body` tags:
```
...
        background-repeat: no-repeat;
        background-position: right 1em top 50%, 0 0;
      }
    </style>

      {% block extra_head %}
      {% endblock %}
    </head>

...

      {% endif %}
    {% endfor %}

  {% block extra_body %}
  {% endblock %}

  </body>
</html>
```

- Deprecate the JS integration function, remove in 2.0
- Move Polaris templates to a subfolder called `polaris` to have a structure like:
  * `templates/polaris/base.html`
  * `templates/polaris/transaction/...`
  * This avoids conflicts the Anchor templates and is a common practice for static files as well
- Update documentation suggesting that if the Anchor needs to add CSS or JS to templates, it can do it by overriding the template.  
For example, let's say the Anchor wants to add styles or scripts to `more_info.html` **only**.  
It could do it by creating a template in `templates/polaris/transaction/more_info.html`:
```
{% extends "polaris/transaction/more_info.html" %}

{% block extra_head %}
[...custom styles...]
[whatever else the Anchor wants at the end of <head>]
{% endblock %}


{% block extra_body %}
<script src='...my custom script...'></script>
[whatever else the Anchor wants at the end of <body>]
{% endblock %}
```

This allows the Anchor to override specific templates instead of loading global JS/CSS in all pages, which is the case with the current integration function.

Issue #317, opened 2020-09-27
Add Support for Django Template Extensions


yuriescl:
When returning HTML from `DepositIntegration.instructions_for_pending_deposit`, I'm having to use inline CSS inside the tags, as I couldn't find an easy way to define custom CSS for Polaris templates.
Polaris has an integration function to define custom JS, it would be great to have something similar for CSS as well.

UPDATE: This issue now is about two things:
- Deprecate integrations that collect data to be used in templates
- Add support for django template extensions

The two integrations that serve the purpose template extensions can do better are:
- JS scripts() integration function
- DepositIntegration.instructions_for_pending_deposit()

`content_for_template` also returns data for templates but its content that must be sent from the backend, not computed or processed on the frontend.

JakeUrban:
The `instructions_for_pending_deposit()` integration needs upgrading. There should be a better interface for providing instructions within the more_info template.

In the immediate future, you can use Polaris' [JS script](https://django-polaris.readthedocs.io/en/stable/sep6_and_sep24/index.html#javascript-integration) integration. Using this integration you can solve your issue a number of ways better described [here](https://dev.to/karataev/set-css-styles-with-javascript-3nl5).

JakeUrban:
Any changes to `instructions_for_pending_deposit()` will be included in a post-v1.1 release. I'm wondering if we should replace the `text/html` content returned with a Django Template object.

The idea is the Polaris dev would load a HTML Template source file into a `Template` django object and return it. Tags can reference CSS style groupings defined from a JS script passed via the JS script integration function mentioned above.

If you don't like the idea of returning `TemplateScript` objects from the JS script integration function and would rather just replace Polaris' CSS styling altogether, see how Polaris allows developers to [override the base CSS stylesheet](https://django-polaris.readthedocs.io/en/stable/sep6_and_sep24/index.html#replacing-polaris-ui-assets) used.

yuriescl:
> I'm wondering if we should replace the text/html content returned with a Django Template object.

Yes. Polaris could support `str` and `Template`. If it's `str`, it treats as plain text and inserts it into `more_info.html` (same as currently). If it's `Template`, it renders the template instead of `more_info.html`.
With that change, I would create a template and inherit it from `more_info.html`.
[...]

PR #323, opened 2020-10-02
Template Extensions

JakeUrban:
resolves stellar/django-polaris#317
resolves stellar/django-polaris#319
resolves stellar/django-polaris#318

Deprecates:
- JS `scripts()` integration function
- `DepositIntegration.instructions_for_pending_deposit()`

Template extensions can provide all of the above functionality. The reason this is better is because there is finally a clear separation between the backend python code and the frontend template logic. Backend servers don't have to return objects containing file paths to JavaScript files so that they can be loaded into a Polaris template.


JakeUrban:
@yuriescl I'd like your thoughts here once this is ready, which should be soon.

JakeUrban:
TODO: add `content_for_template()` support for `error.html` template.

JakeUrban:
@yuriescl @acharb this is ready for review. I'll be added one more call to `content_for_template()` for whenever `error.html` is rendered, that way anchors can override the error pages too.

polaris/polaris/sep24/deposit.py
```diff
@@ -149,6 +149,13 @@ def post_interactive_deposit(request: Request) -> Response:
             )
             or {}
         )
+        if registered_scripts_func != calculate_fee:
```
yuriescl, line 152: @JakeUrban This seems incorrect.
Maybe it should be `if registered_scripts_fund != scripts`?

polaris/polaris/sep24/deposit.py
```diff
@@ -256,6 +263,13 @@ def get_interactive_deposit(request: Request) -> Response:
     elif content is None:
         content = {}
 
+    if registered_scripts_func != calculate_fee:
```
yuriescl, line 266: @JakeUrban Same issue here.
Maybe should be `if registered_scripts_func != scripts`?

polaris/polaris/shared/endpoints.py
```diff
@@ -37,36 +41,38 @@ def more_info(request: Request, sep6: bool = False) -> Response:
...
+        context.update(content)
+    if request_transaction.status == Transaction.STATUS.pending_user_transfer_start:
+        context.update(
+            instructions=rdi.instructions_for_pending_deposit(request_transaction)
         )
 
-    resp_data["scripts"] = registered_scripts_func(content)
-    if content:
-        resp_data.update(content)
+    if registered_scripts_func != calculate_fee:
```
yuriescl, line 62: @JakeUrban Same issue.

polaris/polaris/templates/polaris/base.html
```diff
@@ -230,5 +233,9 @@
       {% endif %}
     {% endfor %}
 
+    {# Rendered via custom extra_head template #}
```
yuriescl, line 236: Not sure what this comment means

polaris/polaris/templates/polaris/deposit.html
```diff
@@ -1,4 +1,4 @@
-{% extends "base.html" %} {% load i18n %}
+{% extends "./base.html" %} {% load i18n %}
```
yuriescl, line 1: Should be extending from `polaris/base.html`. Otherwise the Anchor can't override `polaris/base.html`. I'm not sure about that though.

polaris/polaris/templates/polaris/error.html
```diff
@@ -1,4 +1,4 @@
-{% extends "base.html" %} {% load i18n %}
+{% extends "./base.html" %} {% load i18n %}
```
yuriescl, line 1: See above comment.

polaris/polaris/templates/polaris/more_info.html
```diff
@@ -1,4 +1,4 @@
-{% extends "base.html" %} {% load i18n %}
+{% extends "./base.html" %} {% load i18n %}
```
yuriescl, line 1: See above comment.

polaris/polaris/templates/polaris/withdraw.html
```diff
@@ -1,4 +1,4 @@
-{% extends "base.html" %} {% load i18n %}
+{% extends "./base.html" %} {% load i18n %}
```
yuriescl, line 1: See above comment.

yuriescl:
Since `content_for_template()` returns a context for the template, maybe it would make sense to rename it to `context_for_template()`.

polaris/polaris/sep24/deposit.py
JakeUrban: `scripts` is the keyword argument to the `register_integrations()`. `calculate_fee` is actually the default function Polaris calls if no function was passed to `register_integrations()` `scripts`.

JakeUrban: Good catch, but it should actually be
```
registered_fee_func != calculate_fee
```

polaris/polaris/templates/polaris/deposit.html
JakeUrban: Thats not the case, Django's `extends` keyword resolves relative paths and is able to detect when it is overridden by another app regardless of the hard-coded path. Django compares all paths relative to their respective app's `templates` folder: https://docs.djangoproject.com/en/3.1/ref/templates/api/#django.template.loaders.app_directories.Loader

JakeUrban:
@yuriescl Thanks for the review, I agree `context_for_template()` would have been a better name, but again, breaking changing will have to wait until 2.0. I'm also not sure its worth updating given the names are fairly similar.

polaris/polaris/sep24/deposit.py
JakeUrban: Actually nevermind, you're right completely!

polaris/polaris/templates/polaris/base.html
JakeUrban: When you inspect the HTML rendered in the browser by this template, the JS scripts at the bottom of the body will be either loaded by the JS script integration function or template extensions. I'd like there to be a visual indicator of which source the scripts came from.

yuriescl:
> @yuriescl Thanks for the review, I agree `context_for_template()` would have been a better name, but again, breaking changing will have to wait until 2.0. I'm also not sure its worth updating given the names are fairly similar.

Yes, better to change that in 2.0. I don't see why not update it, 2.0 will introduce breaking changes anyway and it's an easy change for the Anchor.
`context_for_template()` is far more intuitive for a Django dev.

polaris/polaris/templates/polaris/base.html
yuriescl: But shouldn't it be `extra_body` instead of `extra_head` there?

polaris/polaris/sep24/withdraw.py
```diff
@@ -158,7 +159,14 @@ def post_interactive_withdraw(request: Request) -> Response:
             )
             or {}
         )
-        scripts = registered_scripts_func({"form": form, **content})
+        if registered_scripts_func != scripts:
```
acharb, line 162: I haven't tested myself but just wanted to confirm this warning does show up, and also doesn't always show up. Because I see we're assigning [here](https://github.com/stellar/django-polaris/blob/a66d00be0530b680736e168e97b088ef7fa0cc98/polaris/polaris/integrations/javascript.py#L59), and comparing function objects I've found tricky in the past.

Also I think comparing function objects w/ `!=` compares the memory address by default, but maybe better to use `is not` for these, as that always compares addresses. But personal preference here

acharb (approved):
LGTM, 2 comments

example/server/templates/polaris/base.html
```diff
@@ -0,0 +1,23 @@
+{% extends "polaris/base.html" %} {% load static %} {% load i18n %}
+
+{% trans "Confirm Email" as ce %}
+
+{% block extra_body %}
+
+    <script src="https://www.googletagmanager.com/gtag/js?id=UA-53373928-6" async></script>
+    <script src="{% static 'sep24_scripts/google_analytics.js' %}"></script>
+
+    {% if not form is not None and title != ce %}
```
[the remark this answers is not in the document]
JakeUrban: Great catch thanks, it should be `if from is None` not `if not form is None`. I'm not even sure this is correct syntax. It rendered successfully so I guess it is.

polaris/polaris/sep24/withdraw.py
JakeUrban: I agree `is not` is better, changing. That should resolve the object comparison issue. I do agree comparing objects is not the greatest way of determining if a custom function was registered. Maybe we should have a better system in place to do that, but thats for a later date.

Issue #334, opened 2020-10-09
Move distribution account loading into application startup function


yuriescl:
I upgraded to Polaris v1.1.1 and migrated the database. Then, when creating the first asset in the database I got this:
```
Traceback (most recent call last):
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/db/backends/sqlite3/base.py", line 383, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.IntegrityError: UNIQUE constraint failed: polaris_asset.id

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 21, in <module>
    main()
  File "manage.py", line 17, in main
    execute_from_command_line(sys.argv)
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/home/yuri/work/polaris-anchor/app/management/commands/create_asset.py", line 70, in handle
    asset = Asset.objects.create(
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/db/models/query.py", line 422, in create
    obj.save(force_insert=True, using=self.db)
  File "/home/yuri/.virtualenvs/polaris_venv/lib/python3.8/site-packages/django/db/models/base.py", line 743, in save
    self.save_base(using=using, force_insert=force_insert,
[...]

PR #335, opened 2020-10-09
Replace loading distribution account on startup with on-demand in-memory loading

JakeUrban:
resolves stellar/django-polaris#334

Polaris 1.1.0 changed `Asset` initialization to check for whether or not an in-memory cache record existed for the `Asset`'s distribution account. If no record of the asset existed in the cache, Polaris would do two things:

- Make an API call to Horizon for the asset's distribution account
- Make a DB query to save the account's signers and weights from the Horizon response

This is a bug and 1.1.0 as well as 1.1.1 will likely be removed from PyPi. Version `1.1.2` will be released with this change.

### DIFF
--- a/polaris/polaris/models.py
+++ b/polaris/polaris/models.py
@@ -242,6 +242,13 @@ class Asset(TimeStampedModel):
     the account's public key, if present.
     """
 
+    updated_at = models.DateTimeField(default=utc_now)
+    """
+    Updated on every instance ``.save()`` call. Note that multi-object write
+    queries do not call ``.save()``, such as ``Asset.objects.update(...)``.
+    This means such queries must include the a value for ``update_at``.
+    """
+
     objects = models.Manager()
 
     @property
@@ -252,6 +259,14 @@ class Asset(TimeStampedModel):
         if not self.distribution_seed:
             return None
         return Keypair.from_secret(str(self.distribution_seed)).public_key
+
+    def save(self, *args, **kwargs):
+        """
+        Writes any changes made to the model instance in memory to the
+        database. Explictly updates ``updated_at``.
+        """
+        self.updated_at = utc_now()
+        super().save(*args, **kwargs)
 
     class Meta:
         app_label = "polaris"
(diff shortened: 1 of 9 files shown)


JakeUrban:
@yuriescl

yuriescl (approved):
Thanks

polaris/polaris/models.py
```diff
@@ -261,26 +238,64 @@ def __init__(self, *args, **kwargs):
...
+        )
+
+    def load_distribution_account_data(self):
+        from polaris import settings
+
+        if self.code in ASSET_DISTRIBUTION_ACCOUNT_DATA:
+            account_json = ASSET_DISTRIBUTION_ACCOUNT_DATA[self.code]
+        else:
+            account_json = (
+                settings.HORIZON_SERVER.accounts()
```
acharb, line 287: if possible I'd wrap a try-catch around this horizon API call in case it fails. In which case I think it would leave the private fields empty, so it could be loaded later

polaris/polaris/models.py
```diff
@@ -28,36 +27,14 @@
 from stellar_sdk.exceptions import SdkError
 
 
-# This dictionary acts as an in-memory indication of whether or not
-# the Asset object's distribution account fields have been updated
-# since this process started.
-ASSET_DISTRIBUTION_ACCOUNT_LOADED = {}
+# Used for loading the distribution signers data onto an Asset obj
+ASSET_DISTRIBUTION_ACCOUNT_DATA = {}
```
acharb, line 31: nit: `ASSET_DISTRIBUTION_ACCOUNT_DATA` naming is a little confusing to me. It's a dictionary with keys being asset codes, and values being distribution account data?

If so, could we name it to something like `ASSET_TO_DISTRIBUTION_ACCOUNT_MAP` ?

acharb (approved):
LGTM! Left a couple comments

polaris/polaris/models.py
JakeUrban: Thats a better name, thanks.

Issue #284, opened 2020-08-16
Feature Request: Serve Static TOML File


jcooter:
**What problem does your feature solve?**
Adding other SEP-1 token metadata to the Asset object will allow for more robust sep-1 generation, as well as allowing metadata to be used by other django apps and business logic

**What would you like to see?**
SEP-1 defines informational metadata for tokens (e.g. name, image, asset type, etc.). It would be nice to have this information included in the Asset model.

**What alternatives are there?**
Currently, the alternative is to define this manually in a SEP-1 integration (which does not allow other apps to take advantage of the information), or to create a separate db model to contain the metadata with a one-to-one mapping to an Asset (which asks for the user to duplicate boilerplate work)

JakeUrban:
Hi @jcooter, when you say meta data, do you mean this part of SEP-1?
```
# asset with meta info
[[CURRENCIES]]
code="GOAT"
issuer="GD5T6IPRNCKFOHQWT264YPKOZAWUMMZOLZBJ6BNQMUGPWGRLBK3U7ZNP"
display_decimals=2
name="goat share"
desc="1 GOAT token entitles you to a share of revenue from Elkins Goat Farm."
conditions="There will only ever be 10,000 GOAT tokens in existence. We will distribute the revenue share annually on Jan. 15th"
image="https://static.thenounproject.com/png/2292360-200.png"
fixed_number=10000
```
So you'd like the `Asset` model to have `name`, `desc`, `conditions`, etc. ?

jcooter:
Yep, that's what I'm looking for.  I implemented this in a branch off the 0.13.1 release because master was broken at the time.  I can try to put together a PR this weekend.

JakeUrban:
The only hesitation I have is that I intended to squash the migration files used to modify Polaris' DB schema for the 1.0 release, with the assumption that no schema changes would happen between 0.13.1 and 1.0. Can this functionality be slated for a 1.1 release?

JakeUrban:
In the meantime, we should merge your PR into a 1.1 branch

jcooter:
Yeah, that's fine.  I'm using this for a hobby project of mine, so I don't have any pressing need.

JakeUrban:
Hey Joe,

I'm looking at the SEP-1 CURRENCY attribute definitions and I'm beginning to feel as if having a column for each attribute is overkill. Some of these will rarely ever be used, and others may appear to be significant to Polaris but are not, like `status`.

[...]

PR #337, opened 2020-10-13
Static Stellar TOMLs

JakeUrban:
resolves stellar/django-polaris#284

Adds another option for serving the TOML file: serving a static file instead of serializing a dictionary using the python `toml` library.

Note this does not deprecate the original approach. This is likely going to be more straightforward for non-django developers who don't know how static files are found.


JakeUrban:
@yuriescl

JakeUrban:
@jcooter

acharb (approved):
LGTM

Issue #316, opened 2020-09-23
Add support for two-part payments


tomerweller:
Should Polaris support two part payments and sponsored reserves?

https://www.stellar.org/developers-blog/guide-to-protocol-14-prep

JakeUrban:
Polaris already pays for the account's creation if it doesn't exist, but it definitely should add support for two-part-payments.

Zagan202:
## Currently the payment flow is as follows without two-part payments
1. Alice issues a payment of `SRT` to Bob
2. Bob doesn't have a trustline established with `SRT`
3. Alice's transaction fails  
4. Bob establishes a trustline for `SRT`
5. Alice retries to issue a payment of `SRT` to Bob

## The payment flow is as follows with two-part payments
1. Alice issues a payment of `SRT` to Bob
2. Bob doesn't have a trustline established with `SRT`
3. Alices's payment will instead be made as a `claimableBalance`
    - The claimant is Bob
    - The sponsor is Alice such that she can retrieve it if need be
    - **Note:** We could set predicates that state when it should be claimed by
4. Bob establishes a trustline for `SRT`
5. Bob claims the `claimableBalance` (sponsored by Alice) of `SRT`

Zagan202:
One point of change will have to be `create_stellar_deposit`.
On line 150 there's a condition `if created or pending_trust:` 
```
        if created or pending_trust:
            # the account is pending_trust for the asset to be received
            if pending_trust and transaction.status != Transaction.STATUS.pending_trust:
                transaction.status = Transaction.STATUS.pending_trust
                transaction.save()
            return False
```
Instead of exiting here we're going to flag this transaction as a `claimableBalance` and submit it as such 
`return submit_stellar_deposit(transaction, multisig=multisig)`

JakeUrban:
The solution steps shouldn't have the first point, because Polaris shouldn't try to make a payment before checking for the trustline.

Zagan202:
Updating https://github.com/stellar/django-polaris/pull/341 to be compatible with the trustline flow as well as the claimable balance flow.

The claimable balance flow will occur when the key-value claimable_balance_supported: True is given via the POST req payload.

JakeUrban:
@Zagan202 I think I know what you meant but just in case, I would start a new branch instead of building on an already-merged PR.

PR #341, opened 2020-10-16
Two-Part Payment Support

Zagan202:
# Description

Adds two part payments (aka claimableBalances) to the deposit flow.
In the past when accounts attempt a deposit 

### Old Flow
1. The Wallet will initiate a deposit to an account that does not exist yet,
or the user's account won't have a trustline to the asset's issuer account. 
2. Polaris will be run periodically via the ``poll_pending_deposits``
3. Polaris calls ``DepositIntegration.create_channel_account()`` for the transaction record (which creates a temporary, per-deposit-transaction-object Stellar account)
4. Polaris will then mark those created accounts as pending_trust
5. The Wallet will then communicate to the user that the transaction could not be completed until a trust line is established.
   for their own account or the one that was created for them
6. Polaris will call periodically ``check_trustlines`` (a command line tool that periodically checks if the
transactions with this status now have a trustline to the relevant asset). 
7. When the user establishes a trustline Polaris will submit the transaction to the stellar network and call the
``after_deposit`` integration function once its completed.

### Desired Flow

1. The Wallet will initiate a deposit to an account that does not exist yet,
or the user's account won't have a trustline to the asset's issuer account. 
2. Polaris will be run periodically via the ``poll_pending_deposits``
3. Polaris calls ``DepositIntegration.create_channel_account()`` for the transaction record (which creates a temporary, per-deposit-transaction-object Stellar account)
4. Polaris will then mark those created accounts as pending_trust
5. Polaris will be run periodically via the ``poll_pending_deposits``
6. Polaris will create a stellar deposit (``create_stellar_deposit``) create_transaction_envelope where it will build the transaction as a claimableBalance
7. Submit the transaction and mark the status as completed.

Addresses issue: #316 
resolves stellar/stellar-protocol#747

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# How Has This Been Tested?

- [X] `tests/processes/test_create_stellar_deposit.py`


JakeUrban (commented):
Good start, at a high level everything looks good here. 

The key feature I'd like to see is a `Transaction.claimable_balance_id` column that gets populated on successful TX submission.

Other than that, there are some small adjustments to make, and I also think we could add to the documentation to explain when Claimable Balances are used in Polaris.

polaris/polaris/utils.py
```diff
@@ -409,24 +409,39 @@ def get_or_create_transaction_destination_account(
...
-        destination=transaction.stellar_account,
-        asset_code=transaction.asset.code,
-        asset_issuer=transaction.asset.issuer,
-        amount=str(payment_amount),
-        source=transaction.asset.distribution_account,
-    )
+    if claimable:
+        claimant = Claimant(destination=transaction.stellar_account)
+        asset = Asset(code=transaction.asset.code,
+                      issuer=transaction.asset.issuer)
```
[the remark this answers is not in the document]
Zagan202: Oh awesome I wasn't sure if you used a formatter.
Black is definitely my favorite python formatter so I'll re-enable it

polaris/polaris/utils.py
```diff
@@ -409,24 +409,39 @@ def get_or_create_transaction_destination_account(
...
-        network_passphrase=settings.STELLAR_NETWORK_PASSPHRASE,
-        base_fee=base_fee,
-    ).append_payment_op(
-        destination=transaction.stellar_account,
-        asset_code=transaction.asset.code,
-        asset_issuer=transaction.asset.issuer,
-        amount=str(payment_amount),
-        source=transaction.asset.distribution_account,
-    )
+    if claimable:
```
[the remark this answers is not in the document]
Zagan202: Yeah..I really should have thought about that 😅

polaris/polaris/tests/processes/test_create_stellar_deposit.py
```diff
@@ -173,8 +185,5 @@ def test_deposit_stellar_no_trustline(acc1_usd_deposit_transaction_factory):
     deposit = acc1_usd_deposit_transaction_factory()
     deposit.status = Transaction.STATUS.pending_anchor
     deposit.save()
-    assert not create_stellar_deposit(deposit)
-    assert (
-        Transaction.objects.get(id=deposit.id).status
-        == Transaction.STATUS.pending_trust
-    )
+    assert create_stellar_deposit(deposit)
+    assert Transaction.objects.get(id=deposit.id).status == Transaction.STATUS.completed
```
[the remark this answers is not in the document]
Zagan202: Did you want me to charge forward by including the proposed `claimable_balance_id` field in the `GET /Transaction(s)` response?
https://github.com/stellar/stellar-protocol/issues/747#issuecomment-710398419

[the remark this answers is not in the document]
Zagan202: This commit c18d46f extends the Transaction model to include the `claimable_balance_id` field.
However this field will remain empty until https://github.com/StellarCN/py-stellar-base includes the helper Horizon added https://github.com/stellar/go/pull/3122

polaris/polaris/utils.py
```diff
@@ -135,7 +133,9 @@ def create_stellar_deposit(
         transaction.save()
         raise ValueError(transaction.status_message)
 
-    # if we don't know if the destination account exists
+    # if the destination account doesn't exists
```
JakeUrban, line 136: Did you feel the old comment wasn't clear? I think the 2nd condition in your new statement is redundant since thats true for any boolean function parameter that defaults to `False`.

`destination_exists=False` means we don't know if it exists, or rather that we need to check. If you want to rename it to something that makes more sense I'd support that.

polaris/polaris/utils.py
```diff
@@ -420,13 +419,25 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
...
         network_passphrase=settings.STELLAR_NETWORK_PASSPHRASE,
         base_fee=base_fee,
-    ).append_payment_op(
-        destination=transaction.stellar_account,
-        asset_code=transaction.asset.code,
-        asset_issuer=transaction.asset.issuer,
-        amount=str(payment_amount),
-        source=transaction.asset.distribution_account,
     )
+    if transaction.status == Transaction.STATUS.pending_trust:
```
JakeUrban, line 423: So right now, `transaction.status` is only `pending_trust` when `destination_exists=False` and `get_or_create_stellar_destination_account()` returns `pending_trust`. `destination_exists=True` might no longer used since we deleted `check_trustlines`, in which case we should remove it and always check for the account's existence.

polaris/polaris/tests/processes/test_create_stellar_deposit.py
```diff
@@ -173,8 +173,5 @@ def test_deposit_stellar_no_trustline(acc1_usd_deposit_transaction_factory):
     deposit = acc1_usd_deposit_transaction_factory()
     deposit.status = Transaction.STATUS.pending_anchor
     deposit.save()
-    assert not create_stellar_deposit(deposit)
-    assert (
-        Transaction.objects.get(id=deposit.id).status
-        == Transaction.STATUS.pending_trust
-    )
+    assert create_stellar_deposit(deposit)
```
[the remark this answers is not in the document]
JakeUrban: Actually, I don't see a function that extracts the ID from result XDR, did you want to wait off on implementing it?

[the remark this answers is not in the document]
Zagan202: Oh maybe I misunderstood from our sync, I reverted the changes that created the claimable_balance_id column.
I was under the assumption that we'll push that responsibility to the wallet since they can decode and parse the results xdr themselves.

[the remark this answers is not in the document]
Zagan202: I can make an assert here to decode the XDR and check if there was a claimable balance operation made.

polaris/polaris/utils.py
Zagan202: I think I like this path much more. 
Let me make some changes to see if it works as I intend it to.

polaris/polaris/utils.py
Zagan202: I'm planing on removing this logic and comment due to this change discussed here.
https://github.com/stellar/django-polaris/pull/341#discussion_r508100616

polaris/polaris/tests/processes/test_create_stellar_deposit.py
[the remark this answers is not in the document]
JakeUrban: Yea I don't think we should have it returned in the `/transactions` JSON but I think it should still be in the data model.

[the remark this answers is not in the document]
Zagan202: Got it, ok I'll make these changes then.

polaris/polaris/utils.py
```diff
@@ -443,6 +436,16 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
     return builder.build()
 
 
+def get_balanceid(response):
```
JakeUrban, line 439: Nice. 

can we change the name to to match the attribute? `get_balance_id()`? 

I would also add a docstring explaining the return value. Why hex? Is that according to the CAP/SEP? Or would base64 make more sense? Keep in mind Horizon uses base64 for encoding bytes across the board in their API's. The SDK's on the other hand largely expect hex. I would stick to this pattern: base64 for bytes sent or received in requests, hex if will only be used internally. 

Another good way to think about it is, will developers call `get_balance_id()` and immediately call `base64.b64encode()`?

JakeUrban (commented):
Nice, looks good pending my comments.

Also: Did you add black to your git commit hook?

polaris/polaris/utils.py
```diff
@@ -420,18 +412,40 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
...
+        )
     if memo:
         builder.add_memo(memo)
     return builder.build()
 
 
+def get_balanceid(response):
+    claimable_balance_res = response['result_xdr']
+    tr_xdr = TransactionResult.from_xdr(claimable_balance_res)
+    for tr in tr_xdr.result.results:
```
JakeUrban, line 442: Can we return none explictly if we don't find what we're looking for? Also, should this function return none or raise an error? Will developers call this for transactions they know have a claimable balance op or will they use this to determine if one exists?

polaris/polaris/utils.py
```diff
@@ -268,7 +260,7 @@ def submit_stellar_deposit(transaction, multisig=False) -> bool:
             "Stellar transaction failed when submitted to horizon: "
             f"{transaction_result.result.results}"
         )
-
+    transaction.claimable_balance_id = get_balanceid(response) if claimable else None
```
JakeUrban, line 263: If a transaction is `pending_trust`, will the transaction's envelope always include a claimable balance op? Or can it have a payment op? If a transaction can be `pending_trust` and use a payment, `get_balance_id()` will return None, which I guess isn't a problem per se, but a developer looking at this probably believes `get_balance_id()` always returns something.

polaris/polaris/tests/sep24/test_deposit.py
```diff
@@ -24,6 +24,12 @@
     "id": "test_stellar_id",
     "paging_token": "123456789",
 }
+HORIZON_SUCCESS_RESPONSE_CLAIM = {
```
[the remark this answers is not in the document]
Zagan202: I split it originally to not have a `result_xdr` that is unrepresentative of a test that isnt making a claimable balance.
However adding that attribute to `HORIZON_SUCCESS_RESPONSE` will make it cleaner.

polaris/polaris/utils.py
Zagan202: This in order to query and claim claimable balances (on horizon) you need the hex representation of said id.
So to reduce friction I convert this id to hex.

polaris/polaris/utils.py
Zagan202: Technically `get_balance_id()` Returns `None` when the loop has exhausted.

Developers are welcome to use this helper method to parse any responses from `submit_transaction(envelope)` to find if they contain `createClaimableBalanceResult`s.

JakeUrban: I know it _can_ return `None`, but it doesn't _look_ like it does.

polaris/polaris/utils.py
JakeUrban: Oh I see, so _Horizon uses base64 for encoding bytes across the board in their API's._ isn't true. sg

Zagan202:
> Nice, looks good pending my comments.
> 
> Also: Did you add black to your git commit hook?

I did now haha

JakeUrban:
@Zagan202 thanks, can you format the code in a commit and push? I think you just need to change whitespace and black will format the code prior to commit.

Zagan202:
> @Zagan202 thanks, can you format the code in a commit and push? I think you just need to change whitespace and black will format the code prior to commit.

Thats strange everything seemed to check out on my end 
I'm getting `black....................................................................Passed` 
Could I see what your `.pre-commit-config.yaml` is?

Zagan202:
Also ran `pre-commit run --all-files` for good measure.
All the failing files (then formatted) are ones I didn't make changes to.
Would you want me to format and commit those?
I thought its ideal to keep this PR focused on changes that directly impact the issue,
```
pre-commit run --all-files                            
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/integrations/javascript.py
reformatted /Users/lawrence/stellar-dev/django-polaris/example/server/migrations/0003_auto_20200222_0003.py
reformatted /Users/lawrence/stellar-dev/django-polaris/example/server/migrations/0005_auto_20200702_1908.py
reformatted /Users/lawrence/stellar-dev/django-polaris/example/server/migrations/0006_auto_20200615_2046.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/tests/processes/test_watch_transactions.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/settings.py
reformatted /Users/lawrence/stellar-dev/django-polaris/example/server/settings.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/management/commands/watch_transactions.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/shared/endpoints.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/sep31/transactions.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/tests/shared_endpoints/test_transactions.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/tests/sep6/test_deposit.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/migrations/0001_initial.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/tests/sep12/test_customer.py
reformatted /Users/lawrence/stellar-dev/django-polaris/polaris/polaris/tests/sep31/test_send.py
All done! ✨ 🍰 ✨
15 files reformatted, 111 files left unchanged.
```

JakeUrban:
```
(django-polaris) ~/Documents/workspace/stellar/django-polaris (master) $ black --version
black, version 19.10b0
(django-polaris) ~/Documents/workspace/stellar/django-polaris (master) $ cat .git/hooks/pre-commit
echo "Formatting code..."
black polaris
black example
git add .
```

JakeUrban (approved):
LGTM!

JakeUrban:
We can merge this and simply hold off on including it in a release until anchors are expected to use claimable balances.

PR #342, opened 2020-10-24
base.html title block double quotes

yuriescl:
In templates which extend `base.html`, `{% block "title" %}` is not overriding `{% block 'title' %}` from `base.html`, but `{% block 'title' %}` is, so instead of using single quotes when extending, I think it's easier to stick to the double quotes pattern in the base template.

### DIFF
--- a/polaris/polaris/templates/polaris/base.html
+++ b/polaris/polaris/templates/polaris/base.html
@@ -6,7 +6,7 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
-    {% block 'title' %}
+    {% block "title" %}
     <title>{% if title %}{{ title|safe }}{% else %}{% trans "SEP24 reference server" %}{% endif %}</title>
     {% endblock %}
 


JakeUrban:
So after looking at the Django documentation, it appears that the unspoken convention is like so:

[source](https://docs.djangoproject.com/en/2.2/ref/templates/language/)
```
{% extends "base_generic.html" %}

{% block title %}{{ section.title }}{% endblock %}

{% block content %}
<h1>{{ section.title }}</h1>

{% for story in story_list %}
<h2>
  <a href="{{ story.get_absolute_url }}">
    {{ story.headline|upper }}
  </a>
</h2>
<p>{{ story.tease|truncatewords:"100" }}</p>
{% endfor %}
{% endblock %}
```

`extends` values should use double quotes. `block` values not have quotes at all.

Can you adjust the templates to conform to this ^? The changes should be limited to the `polaris/polaris/templates/polaris/...` and `example/server/templates/polaris/...` directories.

JakeUrban:
This is unfortunately a potentially breaking change for some users depending on whether or not they use single or double quotes for `block` or `extends`.

yuriescl:
> From what you described, it sounds like you could just use single quotes for now right?

Yes.

I'm closing this Pull Request, it's probably better to implement this together with the upgrade to Django 3.0

PR #348, opened 2020-11-06
SEP-6: pass all request params to process_sep6_request

### LINKED ISSUE #345: SEP-6 custom fields not passed to process_sep6_request
SEP-6 allows custom fields (defined in `/info`) to be passed to `/deposit` and `/withdraw`, but Polaris does not include those custom fields in the `params` passed to `process_sep6_request`.

yuriescl:
resolves #345


JakeUrban (approved):
Nice, this was even easier than I thought.

PR #351, opened 2020-11-10
more_info.html lang improvement

yuriescl:
When status is `pending_user_transfer_start`, the amount isn't received from the user yet, so it's confusing to have `Amount received` as the label:

image

`Amount to be received` is more appropriate in that case.

### DIFF
--- a/polaris/polaris/templates/polaris/more_info.html
+++ b/polaris/polaris/templates/polaris/more_info.html
@@ -49,7 +49,7 @@
 
     <div class="info-item">
         <div class="info-label">
-            {% if transaction.status == "pending_user_transfer_start" %}
+            {% if transaction.status in ["incomplete", "pending_user_transfer_start"] %}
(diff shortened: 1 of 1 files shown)


polaris/polaris/templates/polaris/more_info.html
```diff
@@ -49,7 +49,11 @@
 
     <div class="info-item">
         <div class="info-label">
-            {% trans "amount received" %}
+            {% if transaction.status == "pending_user_transfer_start" %}
```
JakeUrban, line 52:
```suggestion
            {% if transaction.status in ["incomplete", "pending_user_transfer_start"] %}
```

JakeUrban (commented):
Looks good, one suggestion

JakeUrban:
Will be released in the next version, whether thats 1.2 or 1.1.6

PR #352, opened 2020-11-20
Update SEP-10 implementation to v3.0

JakeUrban:
Upgrades `stellar-sdk` to `3.0.0b3`, which includes breaking changes. The only places we use XDR decoding is in `utils.py` and `watch_transaction.py`.

@overcat , a quick look just to make sure I'm not missing any breaking changes would be create. Polaris doesn't use `.to_xdr_object()` functions. But it does get into the low-level XDR when checking withdrawal transactions in `watch_transactions.py`.


setup.py
```diff
@@ -39,38 +39,38 @@
...
         "django-cors-headers==3.5.0",
         "django-environ==0.4.5",
         "django-model-utils==4.0.0",
-        "djangorestframework==3.12.1",
-        "idna==2.10",
+        "djangorestframework==3.12.2",
+        "idna==2.10; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'",
         "mnemonic==0.19",
-        "multidict==4.7.6",
+        "multidict==5.0.2; python_version >= '3.5'",
```
JakeUrban, line 58: i was concerned about the major version update but it looks like we'll be fine:
https://multidict.readthedocs.io/en/stable/changes.html

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -138,7 +144,7 @@ def process_response(cls, response, account):
         # transaction. This allows anchors to validate the actual amount sent in
         # execute_outgoing_transactions() and handle invalid amounts appropriately.
         transaction.amount_in = round(
-            Decimal(payment_op.amount), transaction.asset.significant_decimals,
+            Decimal(payment_op.amount.int64), transaction.asset.significant_decimals,
```
JakeUrban, line 147: @overcat is this correct? How do I get a string representation of the payment amount?

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -187,11 +193,33 @@ def find_matching_payment_op(
...
+        ]:
+            return False
+        operation = cls._cast_operation(generic_operation)
+        asset = operation.dest_asset.alpha_num4 or operation.dest_asset.alpha_num12
         return (
-            operation.type_code() == Xdr.const.PAYMENT
-            and str(operation.destination) == want_asset.distribution_account
-            and str(operation.asset.code) == want_asset.code
-            and str(operation.asset.issuer) == want_asset.issuer
+            str(operation.destination) == want_asset.distribution_account
```
JakeUrban, line 207: @overcat same here, can I just cast `operation` to get the `G...`/`M...` address? It looks like it returns an object representation:
```
class MuxedAccount:
    ...
    def __str__(self):
        out = []
        out.append(f"type={self.type}")
        out.append(f"ed25519={self.ed25519}") if self.ed25519 is not None else None
        out.append(f"med25519={self.med25519}") if self.med25519 is not None else None
        return f"<MuxedAccount {[', '.join(out)]}>"

```

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -185,13 +185,23 @@ def find_matching_payment_op(
...
+        )
+        asset_code = (
+            asset_num_x.asset_code.asset_code4 or asset_num_x.asset_code.asset_code12
+        ).decode()
         return (
-            operation.type_code() == Xdr.const.PAYMENT
-            and str(operation.destination) == want_asset.distribution_account
-            and str(operation.asset.code) == want_asset.code
-            and str(operation.asset.issuer) == want_asset.issuer
+            str(payment_operation.destination) == want_asset.distribution_account
```
JakeUrban, line 204: @overcat same here, can I just cast `operation` to get the `G...`/`M...` address? It looks like it returns an object representation:
```
class MuxedAccount:
    ...
    def __str__(self):
        out = []
        out.append(f"type={self.type}")
        out.append(f"ed25519={self.ed25519}") if self.ed25519 is not None else None
        out.append(f"med25519={self.med25519}") if self.med25519 is not None else None
        return f"<MuxedAccount {[', '.join(out)]}>"

```

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -185,13 +185,23 @@ def find_matching_payment_op(
...
+            return False
+        payment_operation = PaymentOp.from_xdr(
+            generic_operation.to_xdr_object().to_xdr()
+        )
+        asset_num_x = (
+            payment_operation.asset.alpha_num4 or payment_operation.asset.alpha_num12
+        )
+        asset_code = (
+            asset_num_x.asset_code.asset_code4 or asset_num_x.asset_code.asset_code12
+        ).decode()
```
JakeUrban, line 202: @overcat `asset_code4` and `asset_code12` are `bytes`, will decoding them give me the string representation or an XDR string?

overcat:
@JakeUrban 
Sorry, I'll be on vacation for the next three days. I don't have a computer with me, so I can't deal with these problems. Can you use v2.10.0? It includes support for SEP 10 3.0. I will address these issues when I get back.

JakeUrban:
Oh great, I missed that. Thanks!

Issue #324, opened 2020-10-04
Add pending_external transactions to poll_pending_deposits list


yuriescl:
Polaris [currently calls](https://github.com/stellar/django-polaris/blob/master/polaris/polaris/management/commands/poll_pending_deposits.py#L150) `rri.poll_pending_deposits()` only for transactions with status `pending_user_transfer_start`.
But, some Anchors might wanna set the status as `pending_external` when the funds are received from the user and also require some extra step before returning the transaction from `rri.poll_pending_deposits()`.
For example, if it's a Bitcoin deposit, the user sends the funds to an account (so the funds are received), but the Anchor has to send the funds to its own Bitcoin account, which can take a few minutes, so the correct status would be `pending_external` since we're not exactly waiting for the user to transfer funds.

This is the flow Polaris currently supports:

deposits with status `pending_user_transfer_status` -> passed to `rri.poll_pending_deposits()` -> anchor can maybe add it to the list returned so Polaris can execute the deposit transaction

This is the flow I'd like to have:

1. deposits with status `pending_user_transfer_status` -> passed to `rri.poll_pending_deposits()` -> anchor sets status as `pending_external` but does not add it to the list returned
2. deposits with status `pending_external` -> passed to `rri.poll_pending_deposits()` -> anchor can maybe add it to the list returned so Polaris can execute the deposit transaction

I imagine this would require an extra management command plus an integration function.
This would be similar to `poll_outgoing_transactions` but for deposits.

yuriescl:
This idea is applicable if you assume that `pending_user_transfer_status` is not a suitable status for that intermediate step, which is after receiving the funds from the user and before receiving the funds in the anchor account.

yuriescl:
Or maybe `rri.poll_pending_deposits()` can receive transactions with status `pending_external` as well, and the Anchor can just check the transaction status inside the function to decide how to handle it.

JakeUrban:
> But, some Anchors might wanna set the status as pending_external when the funds are received from the user and also require some extra step before returning the transaction from rri.poll_pending_deposits()

So `pending_external` is actually not the appropriate status here, it is `pending_anchor`. The deposit is in fact received, but you need to do more things before it is ready to be submitted.

So the question becomes, how can we change the `poll_pending_deposits` flow to allow deposit transactions to be `pending_anchor` after `pending_user_transfer_start`.

[...]

PR #354, opened 2020-11-24
Pass 'pending_external' transaction to poll_pending_deposits()

JakeUrban:
resolves stellar/django-polaris#324

Adds `pending_external` `Transaction` objects to the `poll_pending_deposits()` call. This allows anchors to periodically check on transactions they know are en-route but funds have not been received. Prior, anchors were forced to leave these transactions in `pending_user_transfer_start`, which may have confused the user.

Also fixes a test bug where we check the SEP10 domain using `urllib.parse.urlparse`s `hostname` attribute instead of `netloc`.

### DIFF
--- a/polaris/polaris/management/commands/poll_pending_deposits.py
+++ b/polaris/polaris/management/commands/poll_pending_deposits.py
@@ -54,7 +54,7 @@ def check_for_multisig(transaction):
     if transaction.asset.distribution_account_master_signer:
         master_signer = transaction.asset.distribution_account_master_signer
     thresholds = transaction.asset.distribution_account_thresholds
-    if not (master_signer and master_signer["weight"] >= thresholds["med_threshold"]):
+    if not master_signer or master_signer["weight"] < thresholds["med_threshold"]:
         # master account is not sufficient
         transaction.pending_signatures = True
         transaction.status = Transaction.STATUS.pending_anchor
--- a/polaris/polaris/management/commands/poll_pending_deposits.py
+++ b/polaris/polaris/management/commands/poll_pending_deposits.py
@@ -136,14 +136,12 @@ class Command(BaseCommand):
 
     @classmethod
     def execute_deposits(cls):
-        """
-        Right now, execute_deposits assumes all pending deposits are SEP-6 or 24
-        transactions. This may change in the future if Polaris adds support for
-        another SEP that checks for incoming deposits.
-        """
         module = sys.modules[__name__]
         pending_deposits = Transaction.objects.filter(
-            status=Transaction.STATUS.pending_user_transfer_start,
+            status__in=[
+                Transaction.STATUS.pending_user_transfer_start,
+                Transaction.STATUS.pending_external,
+            ],
             kind=Transaction.KIND.deposit,
         )
         try:
--- a/polaris/polaris/tests/auth_test.py
+++ b/polaris/polaris/tests/auth_test.py
@@ -43,7 +43,7 @@ def test_auth_get_account(client):
 
     manage_data_op = transaction_object.operations[0]
     assert manage_data_op.type_code() == Xdr.const.MANAGE_DATA
-    assert manage_data_op.data_name == f"{urlparse(settings.HOST_URL).hostname} auth"
+    assert manage_data_op.data_name == f"{urlparse(settings.HOST_URL).netloc} auth"
     assert len(manage_data_op.data_value) <= 64
     assert len(base64.b64decode(manage_data_op.data_value)) == 48
 
--- a/docs/deployment/index.rst
+++ b/docs/deployment/index.rst
@@ -29,7 +29,7 @@ Polaris periodically queries for transactions in ``pending_external`` and passes
 poll_pending_deposits
 ^^^^^^^^^^^^^^^^^^^^^
 
-Polaris periodically queries for transactions in ``pending_user_transfer_start`` and ``pending_external`` and passes them to the ``RailsIntegration.poll_pending_deposits`` integration function. The anchor is expected to update the transactions' status depending on the if the funds have become available in the anchor's off-chain account or if the funds have been sent but not received. See the ``poll_pending_deposits()`` integration function for more details.
+Polaris periodically queries for transactions in ``pending_user_transfer_start`` and ``pending_external`` and passes them to the ``RailsIntegration.poll_pending_deposits`` integration function. ``Transaction`` objects that represent funds that have been received off-chain should be returned so Polaris can submit them to the Stellar network. See the ``poll_pending_deposits()`` integration function for more details.
 
 check_trustlines
 ^^^^^^^^^^^^^^^^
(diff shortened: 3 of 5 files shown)


polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -54,7 +54,7 @@ def check_for_multisig(transaction):
     if transaction.asset.distribution_account_master_signer:
         master_signer = transaction.asset.distribution_account_master_signer
     thresholds = transaction.asset.distribution_account_thresholds
-    if not (master_signer and master_signer["weight"] >= thresholds["med_threshold"]):
+    if not master_signer or master_signer["weight"] < thresholds["med_threshold"]:
```
JakeUrban, line 57: This logic is equivalent, it just removes the confusing negation.

JakeUrban:
@yuriescl

yuriescl (approved):
Thanks!

polaris/polaris/tests/auth_test.py
```diff
@@ -43,7 +43,7 @@ def test_auth_get_account(client):
 
     manage_data_op = transaction_object.operations[0]
     assert manage_data_op.type_code() == Xdr.const.MANAGE_DATA
-    assert manage_data_op.data_name == f"{urlparse(settings.HOST_URL).hostname} auth"
```
Zagan202, line 46: This test failed when `polaris/.env`'s `HOST_URL` param was `HOST_URL="https://localhost:8000"` as suggested in the documentation https://django-polaris.readthedocs.io/en/stable/tutorials/index.html?highlight=env#specify-environment-variables

However it passed all tests when I removed the port number.
If we want to remain consistent I think we should either change how this pulls the the hostname back to its original behavior
or 
change the documentation to communicate that the port number in `polaris/.env`'s `HOST_URL` is not needed

Zagan202 (changes requested):

polaris/polaris/tests/auth_test.py
JakeUrban: Thats strange, are you sure? I think its the other way around:

`netloc` includes the port number, so when `HOST_URL=http://localhost:8000` it should match, because `SEP10_HOME_DOMAINS` defaults to `urlparse(settings.HOST_URL).netloc`.

`hostname` doesn't have the port number, so when `HOST_URL=http://localhost:8000` is should fail because:
```
>>> urlparse(settings.HOST_URL).hostname
localhost
```
And the polaris settings have it as:
```
>>> urlparse(settings.HOST_URL).netloc
localhost:8000
```

Zagan202: I'm pretty sure.
I got this error when my HOST_URL was "https://localhost:8000"

```
=================================================================== FAILURES ===================================================================
____________________________________________________________ test_auth_get_account _____________________________________________________________

client = <django.test.client.Client object at 0x7f9751115250>

    def test_auth_get_account(client):
        """`GET <auth>` succeeds with a valid TransactionEnvelope XDR."""
        response = client.get(f"{endpoint}?account={STELLAR_ACCOUNT_1}", follow=True)
        content = json.loads(response.content)
        assert content["network_passphrase"] == "Test SDF Network ; September 2015"
        assert content["transaction"]
    
        envelope_xdr = content["transaction"]
        envelope_object = TransactionEnvelope.from_xdr(
            envelope_xdr, network_passphrase=settings.STELLAR_NETWORK_PASSPHRASE
        )
        transaction_object = envelope_object.transaction
        assert transaction_object.sequence == 0
        assert len(transaction_object.operations) == 1
    
        manage_data_op = transaction_object.operations[0]
        assert manage_data_op.type_code() == Xdr.const.MANAGE_DATA
>       assert manage_data_op.data_name == f"{urlparse(settings.HOST_URL).netloc} auth"
E       AssertionError: assert 'localhost auth' == 'localhost:8000 auth'
E         - localhost:8000 auth
E         ?          -----
E         + localhost auth

polaris/polaris/tests/auth_test.py:46: AssertionError
------------------------------------------------------------- Captured stderr call -------------------------------------------------------------
2020-11-25 16:29:33,988 - INFO - polaris.sep10.views: Returning SEP-10 challenge for account GAIRMDK7VDAXKXCX54UQ7WQUXZVITPBBYH33ADXQIADMDTDVJMQGBQ6V
-------------------------------------------------------------- Captured log call ---------------------------------------------------------------
INFO     polaris.sep10.views:views.py:72 polaris.sep10.views: Returning SEP-10 challenge for account GAIRMDK7VDAXKXCX54UQ7WQUXZVITPBBYH33ADXQIADMDTDVJMQGBQ6V
=============================================================== warnings summary ===============================================================
../../.virtualenvs/django-polaris-vKUchHve/lib/python3.7/site-packages/urllib3/util/retry.py:255
  /Users/lawrence/.virtualenvs/django-polaris-vKUchHve/lib/python3.7/site-packages/urllib3/util/retry.py:255: DeprecationWarning: Using 'method_whitelist' with Retry is deprecated and will be removed in v2.0. Use 'allowed_methods' instead
    DeprecationWarning,

-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========================================================== short test summary info ============================================================
FAILED polaris/polaris/tests/auth_test.py::test_auth_get_account - AssertionError: assert 'localhost auth' == 'localhost:8000 auth'
```

JakeUrban: This branch has the following definition in `polaris/polaris/settings.py`
```
SEP10_HOME_DOMAINS = env_or_settings("SEP10_HOME_DOMAINS", list=True) or [
    urlparse(HOST_URL).netloc
]
```
So it doesn't make sense that it would fail to match:
```
f"{urlparse(settings.HOST_URL).netloc} auth"
```

JakeUrban: Tests are passing in this PR, why is your branch different?

Zagan202: If I'm understanding this correctly youre referring to this piece of code.
```
HOST_URL = env_or_settings("HOST_URL")
if not HOST_URL.startswith("http"):
    raise ImproperlyConfigured("HOST_URL must include a protocol (http or https)")

SEP10_HOME_DOMAINS = env_or_settings("SEP10_HOME_DOMAINS", list=True) or [
    urlparse(HOST_URL).netloc
]
```
Then that means `f"{urlparse(settings.HOST_URL).netloc} auth"` is looking at `env_or_settings("HOST_URL")` correct?

Zagan202: If the intention was to pull the `SEP10_HOME_DOMAINS` then we would need to change this to `f"{urlparse(settings.SEP10_HOME_DOMAINS).netloc} auth"` but that poses different problems if its a list of home domains if Im not mistaken.

Zagan202: You could do something like this if we were looking at the `SEP10_HOME_DOMAINS`
But its very inelegant 
`f"{urlparse(settings.SEP10_HOME_DOMAINS[0]).path} auth"`

Zagan202 (approved):

Issue #355, opened 2020-11-24
watch_transactions only supports payment ops, not path payments


JakeUrban:
Unfortunately I've discovered a significant feature missing from Polaris. The `watch_transactions` streaming process only detects and matches transactions with Payment ops, but it should also detect and match Path Payments ops. 

This is likely not a problem for SEP-6 & 24, but is definitely a problem for SEP-31. SEP-31 senders are most often in a different country and use a different currency than the SEP-31 receiver. This means Polaris supports SEP-31 only if the sender is able to make a payment operation instead of a path payment.

This behavior is out of compliance with SEP-31 and will be fixed in the next release.

I'm not going to release a "patch" because the missing feature doesn't _break_ anything that currently exists, its a new feature that should have been added when SEP-31 support was added in general.

PR #356, opened 2020-11-25
Match Path Payments from Clients

JakeUrban:
resolves stellar/django-polaris#355

The issue describes the problem better, but basically we need to support matching streamed payments in all forms, not just via payment operations. So I added support for matching path payment ops, both strict send and strict receive.

### DIFF
--- a/polaris/polaris/management/commands/watch_transactions.py
+++ b/polaris/polaris/management/commands/watch_transactions.py
@@ -1,17 +1,19 @@
 """This module defines custom management commands for the app admin."""
 import asyncio
-from typing import Dict, Optional
+from typing import Dict, Optional, Union, List
 from decimal import Decimal
 
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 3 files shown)


JakeUrban:
Still updating the docs but the code is done

JakeUrban:
Actually, I didn't need to change much on the docs because it is expected that path payments work, since path payments are the basis for SEP-31 cross-border payments.

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -95,16 +107,13 @@ def process_response(cls, response, account):
             return
 
         try:
-            stellar_transaction_id = response["id"]
+            _ = response["id"]
```
Zagan202, line 110: Could this be removed if we're not using it?

Zagan202 (approved):
Besides 1 comment LGTM

PR #358, opened 2020-11-25
Claimable balance: Attempt 2

Zagan202:
# Description

Submits deposit as a claimable balance if the wallet supports claimable balances via sending "claimable_balance_supported=True" in their POST request body.
and the wallet user does not trust the asset their depositing.

If claimable_balance_supported=False or they do not sent that value in `POST /transactions/deposit/interactive` request body and the wallet user does not trust the asset their depositing
then the deposit flow is blocked via pending_trust. 
At which point the wallet user must first create a trust line to that asset in order to complete the deposit.

Addresses issue: #316 

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

# How Has This Been Tested?

- [X] pytest -c polaris/pytest.ini  
    - 1 new test `test_deposit_stellar_no_trustline_with_claimable_bal`
- [X] Submit a deposit from a wallet to Polaris (Where the user doesn't trust the asset.)

The following are steps in order to test this behavior on your end.
1. Checkout the PR branch `git checkout 316-claimable-balance-attempt2` 
1. Checkout the stellar-demo-wallet branch from my fork https://github.com/Zagan202/stellar-demo-wallet/tree/deposit-untrusted-TEST-asset `git checkout deposit-untrusted-TEST-asset`
    - This is a janky test branch that allows the user to deposit offchain value for an asset named SRT
    - Balance should look like this 
Screen Shot 2020-11-30 at 11 26 54 AM
1. Start the wallet `npm start`
1. Build and run Polaris (server and poll_pending_deposits)
   - docker-compose build
   - Execute in one terminal plane `docker-compose up server`
   - Execute in another terminal plane `docker-compose up poll_pending_deposits`
1. Verify that your SRT asset is setup properly 
   - Checkout your `http://localhost:8000/.well-known/stellar.toml`
   - Checkout `http://localhost:8000/adminpolaris/asset/`
1. Login to this [account](http://localhost:3333/?secretKey=SD3XK7YPIJMVUCAFIP34URDZNG3A4QUWCGW6TB4I2JHRQ7DVZCU3G4R2) 
1. Execute a SRT deposit
   - It give you the interactive window then you'll see a claimable balance available for whatever you deposited


polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -220,6 +231,14 @@ def execute_deposits(cls):
         for t in ready_multisig_transactions:
             cls.execute_deposit(t)
 
+        ready_claimable_transactions = Transaction.objects.filter(
+            kind=Transaction.KIND.deposit,
+            status=Transaction.STATUS.pending_anchor_claimable_start,
```
JakeUrban, line 236: As I said in another comment, we can't add a `status` value for our own internal processing, `status` must match the SEPs. If you need to be able to determine if a transaction should be sent using CBs, you need another method of doing so.

Ideally there is no change to the data model and Polaris just figures it out at runtime. Or, we could add an additional column, similar to our `pending_signatures` column, which Polaris uses to determine if a transaction needs additional signers before submitting.

Don't pick the solution that you can implement the easiest, pick the best solution in terms of design and efficiency.

polaris/polaris/migrations/0004_auto_20201120_1541.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.16 on 2020-11-20 15:41
```
JakeUrban, line 1: This migration file is out of date compared to the changes I see. There is no update of the `status` column but you've added a new value to it. 

Since that value will be removed from `status` this looks good now but always make sure your migrations match the current definition of the model.

polaris/polaris/models.py
```diff
@@ -336,6 +336,9 @@ class Transaction(models.Model):
         "pending_trust": _("waiting for a trustline to be established"),
         "pending_user": _("waiting on user action"),
         "pending_user_transfer_start": _("waiting on the user to transfer funds"),
+        "pending_anchor_claimable_start": _(
+            "No trustline, Processing claimable balance"
+        ),
```
JakeUrban, line 341: This will need to be removed.

polaris/polaris/models.py
```diff
@@ -414,6 +417,10 @@ class Transaction(models.Model):
     
         the user has not yet initiated their transfer to the anchor. This is the 
         necessary first step in any deposit or withdrawal flow.
+    * **pending_anchor_claimable_start**
```
JakeUrban, line 420: Same here

polaris/polaris/models.py
```diff
@@ -579,6 +586,15 @@ class Transaction(models.Model):
...
     add itself.
     """
+    claimable_balance_supported = models.BooleanField(default=False)
+    """
+    claimable_balance_supported is a boolean to indicate if the wallet supports the SEP24
+    requirements for handeling claimable balance deposits.
+    """
+    claimable_balance_id = models.TextField(null=True, blank=True)
+    """
+    ClaimableBalanceID is a required parameter when claiming a Claimable Balance operation
```
JakeUrban, line 596:
```suggestion
    The ID of the claimable balance used to send funds to the user. This column will be ``None`` if ``claimable_balance_supported`` is ``False`` or if the transaction has not yet been submitted to the Stellar network.
```

polaris/polaris/sep24/deposit.py
```diff
@@ -313,6 +313,9 @@ def deposit(account: str, request: Request) -> Response:
     stellar_account = request.POST.get("account")
     lang = request.POST.get("lang")
     sep9_fields = extract_sep9_fields(request.POST)
+    claimable_balance_supported = bool(
+        request.POST.get("claimable_balance_supported") == "true"
```
JakeUrban, line 317: Lets make sure our processing of the request parameter is in-line with the changes outlined in the SEP. Does the request value _need_ to be `true`? Can it be `1`, `True`, or `TRUE`? Gotta be careful with these string comparisons.

polaris/polaris/sep24/deposit.py
```diff
@@ -368,6 +371,7 @@ def deposit(account: str, request: Request) -> Response:
         status=Transaction.STATUS.incomplete,
         to_address=account,
         protocol=Transaction.PROTOCOL.sep24,
+        claimable_balance_supported=claimable_balance_supported,
```
JakeUrban, line 374: Ok, so it seems we already have a column that indicates if the transaction should be submitted via CBs, why do we need the new `status` value?

polaris/polaris/tests/processes/test_create_stellar_deposit.py
```diff
@@ -169,6 +169,11 @@ def test_deposit_stellar_no_trustline(acc1_usd_deposit_transaction_factory):
     `create_stellar_deposit` sets the transaction with the provided `transaction_id` to
     status `pending_trust` if the provided transaction's Stellar account has no trustline
     for its asset. (We assume the asset's issuer is the server Stellar account.)
+
+    This is the flow when a wallet/client submits a deposit without the sending
+    "claimable_balance_supported=True" in their POST request body.
+    Such that the deposit transaction is blocked
+    until the Wallet user creates a trustline to that asset
```
JakeUrban, line 176: Lets assert that `claimable_balance_supported` is `False` too

polaris/polaris/tests/processes/test_create_stellar_deposit.py
```diff
@@ -178,3 +183,44 @@ def test_deposit_stellar_no_trustline(acc1_usd_deposit_transaction_factory):
         Transaction.objects.get(id=deposit.id).status
         == Transaction.STATUS.pending_trust
     )
+
+
+mock_server_no_trust_account = Mock(
```
JakeUrban, line 188: Whats the difference between this object and the one at L64?

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -220,6 +231,14 @@ def execute_deposits(cls):
...
+        )
+        for t in ready_claimable_transactions:
+            logger.info(t.id)
```
JakeUrban, line 239: Remove the log statement

polaris/polaris/sep24/deposit.py
JakeUrban: Oh I see, its because we shouldn't use CBs if the account has a trust-line. Ideally, we figure out if it has a trust-line and construct/submit the appropriate transaction without saving anything extra to the transaction object.

polaris/polaris/tests/processes/test_create_stellar_deposit.py
```diff
@@ -178,3 +183,44 @@ def test_deposit_stellar_no_trustline(acc1_usd_deposit_transaction_factory):
...
+    """
+    This is the flow when a wallet/client submits a deposit and the sends
+    "claimable_balance_supported=True" in their POST request body.
+    Such that the deposit transaction is can continue as a
+    claimable balance operation rather than a payment operation
+
+    `execute_deposits` from `poll_pending_deposits.py` sets the transaction.status of the given transaction
+    pulled from the db to `pending_anchor_claimable_start`
+    if the provided transaction's Stellar account has no trustline
+    and sees claimable_balance_supported set to True.
```
JakeUrban, line 217: I think this test will need to be adjusted after the changes I've requested.

polaris/polaris/tests/sep24/test_deposit.py
```diff
@@ -23,6 +23,7 @@
     "successful": True,
     "id": "test_stellar_id",
     "paging_token": "123456789",
+    "result_xdr": "AAAAAAAAAGQAAAAAAAAAAQAAAAAAAAAOAAAAAAAAAAAyBzvi/vP0Bih6bAqRNkiutMVUkW1S+WtuITJAA2LOjgAAAAA=",
```
JakeUrban, line 26: I'd prefer to have a difference success response object used for CBs, since most success responses will have `result_xdr`s for payment ops.

polaris/polaris/utils.py
```diff
@@ -109,19 +109,32 @@ def verify_valid_asset_operation(
...
     """
     Create and submit the Stellar transaction for the deposit.
-
-    The Transaction can be either `pending_anchor` if the task is called
-    from `poll_pending_deposits()` or `pending_trust` if called from the
-    `check_trustlines()`.
+    The Transaction `pending_anchor` when the task is called from `poll_pending_deposits()`.
+    The Transaction `pending_trust` when the task is called from `check_trustlines()`.
+    The Transaction `pending_anchor_claimable_start` when the task is called from `poll_pending_deposits()`
+        and the Transaction is marked as pending_anchor_claimable_start.
```
JakeUrban, line 118: This is a bit redundant, seems like this case can be removed and the transaction should be `pending_anchor` at this point. Then in this function, we determine if we need to use CBs based on `claimable_balance_supported` and whether or not there is a trust-line, which we also determine in this function.

polaris/polaris/utils.py
```diff
@@ -420,18 +435,60 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
...
+        response: the response from horizon
+
+    :return:
+        hex representation of the balanceID
+        or
+        None (if no createClaimableBalanceResult operation is found)
+    """
+    result_xdr = response["result_xdr"]
+    tr_xdr = TransactionResult.from_xdr(result_xdr)
+    for tr in tr_xdr.result.results:
```
JakeUrban, line 483: For the sake of readability I would change the `tr` loop variable to `op_result` or `result`, that way `tr.tr` isn't confusing.

polaris/polaris/utils.py
```diff
@@ -420,18 +435,60 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
...
+        if hasattr(tr.tr, "createClaimableBalanceResult"):
+            cbr_xdr = base64.b64decode(
+                tr.tr.createClaimableBalanceResult.balanceID.to_xdr()
+            )
+            return cbr_xdr.hex()
```
JakeUrban, line 488: I'm not familiar with how `balanceID` should be decoded, but this seems off. Maybe I'm wrong, but are you sure `XDR` --> `hex` is correct? I suspect its `XDR` --> `bytes` --> `hex`.

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -40,13 +41,16 @@ def execute_deposit(transaction: Transaction) -> bool:
             f"Unexpected transaction status: {transaction.status}, expecting "
             f"{' or '.join(valid_statuses)}."
         )
-    if transaction.status != Transaction.STATUS.pending_anchor:
+    if (
+        transaction.status != Transaction.STATUS.pending_anchor
+        and transaction.status != Transaction.STATUS.pending_anchor_claimable_start
```
JakeUrban, line 46: This is an indication that `pending_anchor_claimable_start` is not needed, `pending_anchor` suffices

JakeUrban (changes requested):
After mulling over my comments and potential solutions, let me know if you have questions!

polaris/polaris/management/commands/check_trustlines.py
```diff
@@ -124,7 +124,21 @@ def check_trustlines():
...
+                    if create_stellar_deposit(transaction):
+                        transaction.refresh_from_db()
+                        try:
+                            rdi.after_deposit(transaction)
+                        except Exception:
+                            logger.exception(
+                                "An unexpected error was raised from "
+                                "after_deposit() in check_trustlines"
+                            )
+            for transaction in transactions:
```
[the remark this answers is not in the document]
Zagan202: oh yikes I reverted this commit but didnt push it.
Stand by this logic is obsolete, theres no need to change `check_trustlines` with the current working deposit logic

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -32,6 +32,7 @@ def execute_deposit(transaction: Transaction) -> bool:
     valid_statuses = [
         Transaction.STATUS.pending_user_transfer_start,
         Transaction.STATUS.pending_anchor,
+        Transaction.STATUS.pending_anchor_claimable_start,
```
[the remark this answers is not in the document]
Zagan202: Yes it's my intention to add it to the SEP.
I wanted to communicate a more accurate status of what type of operation is occurring to the wallet.
As well as use it as flag to properly route a transaction to the claimable balance flow.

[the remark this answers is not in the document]
JakeUrban: I'm not convinced its necessary. Wallets would be able to see a CB was sent by checking the `claimable_balance_id` attribute on the `/transactions` endpoint, or by querying horizon for the transaction and figuring it out themselves. Why do we need `pending_anchor_claimable_start` if `pending_anchor` works?

polaris/polaris/tests/sep24/test_deposit.py
Zagan202: I added the `results_xdr` key value in response to a comment you made in the last CB attempt
https://github.com/stellar/django-polaris/pull/341#discussion_r509778548

Would you prefer to keep it how it is currently or have a a difference success response object?

polaris/polaris/management/commands/poll_pending_deposits.py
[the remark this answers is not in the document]
JakeUrban: We don't have a `pending_anchor_payment_start`

polaris/polaris/utils.py
Zagan202: Good idea!

polaris/polaris/management/commands/poll_pending_deposits.py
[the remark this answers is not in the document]
Zagan202: Thats true, let me go ahead and take that out and make the necessary adjustments.

polaris/polaris/utils.py
Zagan202: So I updated this logic so it actually is still pending_trust however since `claimable_balance_supported` is True we'll "do the right thing" and craft a claimable balance operation and submit the transaction.
I've detailed the flow in the docstring

polaris/polaris/migrations/0004_auto_20201120_1541.py
Zagan202: Removed the `pending_anchor_claimable_start` status.

polaris/polaris/sep24/deposit.py
Zagan202: When the wallet sends boolean `true` in the POST req body from its end Python reads it as string `"true"`.
So I wanted to normalize it to a boolean on Polaris' end.

I can make it a bit more forgiving by checking if its returned as `1`, `True`, or `TRUE` (the latter two via python's [lower](https://www.kite.com/python/docs/builtins.str.lower))

polaris/polaris/sep24/deposit.py
Zagan202: With the new logic the transaction obj could be `claimable_balance_supported` and have a trustline established.
If thats the case then Polaris does the right thing and submits the transaction as a payment op.

polaris/polaris/tests/processes/test_create_stellar_deposit.py
Zagan202: Done and done.

polaris/polaris/tests/sep24/test_deposit.py
Zagan202: I decided to go ahead and make a `HORIZON_SUCCESS_RESPONSE_CLAIM` object to avoid `result_xdr` confusion.

polaris/polaris/sep24/deposit.py
JakeUrban: I believe JSON has native `true/false` values, so I would think Django would parse those as native Python booleans.

But that was kinda beside my point. My point was that the parsing logic needs to match the SEP. So if the SEP specifically mentions `claimable_balance_supported` _must_ have a value of `"true"`, then what you have is good 👍

polaris/polaris/management/commands/poll_pending_deposits.py
Zagan202: Reverted the `pending_anchor_claimable_start` status changes.

polaris/polaris/tests/sep24/test_deposit.py
JakeUrban: Lol welp, that was dumb of me. I'll leave this to you.

polaris/polaris/tests/processes/test_create_stellar_deposit.py
```diff
@@ -169,12 +172,59 @@ def test_deposit_stellar_no_trustline(acc1_usd_deposit_transaction_factory):
...
+
+    This is the flow when a wallet/client submits a deposit without the sending
+    "claimable_balance_supported=True" in their POST request body.
+    or sending "claimable_balance_supported=False" in their POST request body.
+    Such that the deposit transaction is blocked
+    until the Wallet user creates a trustline to that asset
     """
     deposit = acc1_usd_deposit_transaction_factory()
-    deposit.status = Transaction.STATUS.pending_anchor
+    deposit.status = Transaction.STATUS.pending_trust
```
JakeUrban, line 183: I actually wouldn't make the transaction's status `pending_trust` because that has historically meant the wallet needs to establish the trustline and the anchor is blocked. But `pending_anchor` indicates that the anchor is the one who has things to do and the wallet can sit tight.

polaris/polaris/utils.py
Zagan202: I am confident that this is how we can currently decode the balanceID from the xdr.

Currently the python SDK's `to_xdr` only returns the result in `base64` representation,
whereas the js SDK gives the option to return this value in its hexadecimal representation.
The hex representation is needed to actually query horizon for the claimable balance.

I'm more or less emulation what @abuiles did [here](https://repl.it/@abuiles/ScientificFriendlyExpertise#index.js)

polaris/polaris/tests/sep24/test_deposit.py
Zagan202: Lol, nah its chill I'll keep them separate.
It's cleaner this way imo

polaris/polaris/tests/processes/test_create_stellar_deposit.py
Zagan202: Ah ok I can revert this then.
I think the docstring threw me off a bit.

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -203,7 +212,10 @@ def execute_deposits(cls):
                 ):
                     transaction.status = Transaction.STATUS.pending_trust
                     transaction.save()
-                continue
+                if not transaction.claimable_balance_supported:
```
JakeUrban, line 215: If `transaction.claimable_balance_supported` is `True`, we don't want to mark the transaction as `pending_trust`. `pending_trust` tells the wallet "the anchor is blocked". But Polaris isn't blocked if the wallet supports CBs.

polaris/polaris/sep24/deposit.py
JakeUrban: Ok, I took a look at the SEP PR for CBs and it looks like this is how the parameter is defined:
Screen Shot 2020-12-01 at 12 56 09 PM

I think you need to do some digging into what a boolean looks like in a POST request. Does it come in as a string, `"true"`, like you're checking for now? Or does it come in as a native python boolean, `True`?

polaris/polaris/tests/processes/test_create_stellar_deposit.py
```diff
@@ -178,3 +187,44 @@ def test_deposit_stellar_no_trustline(acc1_usd_deposit_transaction_factory):
...
+    claimable balance operation rather than a payment operation
+
+    `execute_deposits` from `poll_pending_deposits.py`
+    if the provided transaction's Stellar account has no trustline
+    and sees claimable_balance_supported set to True.
+    it will create_stellar_deposit()
+    where it wil complete the deposit flow as a claimable balance
+    """
+    deposit = acc1_usd_deposit_transaction_factory()
+    deposit.status = Transaction.STATUS.pending_trust
```
JakeUrban, line 225: Not `pending_trust`, `pending_anchor`.

polaris/polaris/management/commands/poll_pending_deposits.py
JakeUrban: I think you need to check for `claimable_balance_supported` prior to marking the transaction as `pending_trust`.

polaris/polaris/utils.py
```diff
@@ -109,15 +109,42 @@ def verify_valid_asset_operation(
...
+    *The Transaction `pending_trust` when the task is called from `check_trustlines()`.*
+
+    If called from check_trustlines() this means the following flow occurred
+    - The deposit transaction was blocked from sending a payment since the account
+      didn't have a trustline
+      - The account also issued a deposit from a Wallet that did not provide a
+        "claimable_balance_supported" value POST req body (or stated it as False).
+    - The Transaction was in pending_trust limbo until the account set up a trustline
+    - Trustline is established for the asset and create_stellar_deposit is called from
+      check_trustlines.py 127,1: if create_stellar_deposit(transaction):
```
JakeUrban, line 125: Don't reference specific code lines in comments
```suggestion
      check_trustlines()
```

polaris/polaris/utils.py
```diff
@@ -420,18 +445,68 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
...
+        None (if no createClaimableBalanceResult operation is found)
+    """
+    result_xdr = response["result_xdr"]
+    tr_xdr = TransactionResult.from_xdr(result_xdr)
+    for op_result in tr_xdr.result.results:
+        if hasattr(op_result.tr, "createClaimableBalanceResult"):
+            cbr_xdr = base64.b64decode(
+                op_result.tr.createClaimableBalanceResult.balanceID.to_xdr()
+            )
+            return cbr_xdr.hex()
```
JakeUrban, line 506: Nice. This is what I thought needed to change. Its not `XDR --> Hex` its `XDR --> Raw Bytes --> Hex`.

polaris/polaris/utils.py
```diff
@@ -420,18 +445,68 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
...
+    The hex representation of the balanceID is important because its the
+    representation required to query and claim claimableBalances.
+
+    :param
+        response: the response from horizon
+
+    :return:
+        hex representation of the balanceID
+        or
+        None (if no createClaimableBalanceResult operation is found)
```
JakeUrban, line 497: Thanks for adding documentation on parameter and return types, but I think the best way to do this is using Python's `typing` module. I added a suggestion to show you what I mean. The docstring above `:param` sufficiently explains whats happening, no `:param` or `:return:` necessary.

polaris/polaris/utils.py
```diff
@@ -420,18 +445,68 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
...
+            asset_issuer=transaction.asset.issuer,
+            amount=str(payment_amount),
+            source=transaction.asset.distribution_account,
+        )
     if memo:
         builder.add_memo(memo)
     return builder.build()
 
 
+def get_balance_id(response):
```
JakeUrban, line 479:
```suggestion
def get_balance_id(response: Dict) -> Optional[str]:
```

polaris/polaris/utils.py
```diff
@@ -268,7 +291,9 @@ def submit_stellar_deposit(transaction, multisig=False) -> bool:
             "Stellar transaction failed when submitted to horizon: "
             f"{transaction_result.result.results}"
         )
-
+    transaction.claimable_balance_id = (
+        get_balance_id(response) if transaction.claimable_balance_supported else None
+    )
```
JakeUrban, line 296: No need to optimize for one-liners, `claimable_balance_id` is already `None`
```suggestion
    if transaction.claimable_balance_supported:
        transaction.claimable_balance_id = get_balance_id(response)
```

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -32,6 +32,7 @@ def execute_deposit(transaction: Transaction) -> bool:
     valid_statuses = [
         Transaction.STATUS.pending_user_transfer_start,
         Transaction.STATUS.pending_anchor,
+        Transaction.STATUS.pending_trust,
```
JakeUrban, line 35: Transaction shouldn't be `pending_trust` when this is called. If CBs are supported by the wallet, then its the transaction should be `pending_anchor` because we don't need to wait for the trustline.

polaris/polaris/utils.py
```diff
@@ -135,24 +162,20 @@ def create_stellar_deposit(
...
+        )
+    except RuntimeError as e:
+        transaction.status = Transaction.STATUS.error
+        transaction.status_message = str(e)
+        transaction.save()
+        logger.error(transaction.status_message)
+        return False
+    if created or pending_trust:
+        if pending_trust and transaction.status != Transaction.STATUS.pending_trust:
+            transaction.status = Transaction.STATUS.pending_trust
```
JakeUrban, line 177: So it looks like the change here is we're no longer doing `return False`. But if the account doesn't have a trust-line and `transaction.claimable_balance_supported` is `False`, we still want to `return False`.

polaris/polaris/utils.py
```diff
@@ -420,18 +445,68 @@ def create_transaction_envelope(transaction, source_account) -> TransactionEnvel
...
-        destination=transaction.stellar_account,
-        asset_code=transaction.asset.code,
-        asset_issuer=transaction.asset.issuer,
-        amount=str(payment_amount),
-        source=transaction.asset.distribution_account,
     )
+    if (
+        transaction.claimable_balance_supported
+        and transaction.status == Transaction.STATUS.pending_trust
+    ):
```
JakeUrban, line 452: Instead of checking `status`, we'll need to check if the destination account has a trust-line.

JakeUrban (changes requested):

polaris/polaris/utils.py
JakeUrban: There are a couple other places below too

Zagan202:
Hey @JakeUrban I fixed a bug in https://github.com/stellar/django-polaris/pull/358/commits/51ab11a92f2bcda74c409ea333fcfd3e69c859ef where we were not triggering `execute_deposit` 
When the `continue` statement is called we do not move to the next executable line within that loop iteration.
Instead we continue to the next iteration.

This was a problem because if we're on the `nth` transaction and `transaction.claimable_balance_supported` we then `check_for_multisig` and then we go to transaction `n+1` (`execute_deposit` is not called) 
or  we're on the `nth` transaction and not `transaction.claimable_balance_supported` and not  `created or pending_trust` then `check_for_multisig` and then we go to transaction `n+1` (`execute_deposit` is not called)

We only want to `continue` to `n+1` without calling `execute_deposit` when `created or pending_trust`.

Correct me if I'm wrong.

polaris/polaris/utils.py
```diff
@@ -113,9 +113,39 @@ def create_stellar_deposit(transaction: Transaction) -> bool:
...
+    False otherwise. (sep24/deposit.py)
+    - Then `poll_pending_deposits()` will do the following
+        - Parse the deposit Transactions that are pending_user_transfer_start
+        - Since transaction.claimable_balance_supported is True
+            call `create_stellar_deposit` from `execute_deposit`
+            where we then craft and submit the transaction as a claimable balance
+        - In `create_transaction_envelope`
+            if claimable_balance_supported and `has_trustline` then
+        deposit is claimablebalance op
+            else deposit is payment op
```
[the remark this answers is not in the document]
JakeUrban: Rather, the docstring should be a short explanation of what the function does. Something like: `Calls submit_stellar_deposit() after performing final status and signature checks. Places transactions in an error status if submit_stellar_deposit() raises an exception.`

JakeUrban (approved):
LGTM! I think the comments/docstrings can be adjusted but otherwise is good to go.

JakeUrban (approved):
Looks good to merge!

Issue #359, opened 2020-12-04
Refactor the `poll_pending_deposit` flow


Zagan202:
Currently the `poll_pending_deposit` process has redundant and cyclical paths.
## Current Flow
<img width="820" alt="Screen Shot 2020-12-04 at 9 02 32 AM" src="https://user-images.githubusercontent.com/5233187/101201685-231a6b80-3636-11eb-86a4-eb501918d79b.png">

## Desired Flow
<img width="812" alt="Screen Shot 2020-12-04 at 9 07 29 AM" src="https://user-images.githubusercontent.com/5233187/101201686-23b30200-3636-11eb-9213-5fe74a4d7d1d.png">

PR #360, opened 2020-12-10
Removed path for the create_stellar_deposit to ""get_or_ create"

Zagan202:
# Description

Decouples the creation of an account from the create deposit method

Addresses issue: #359 

## Type of change

- [X] Bug fix / Refactor (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] `pytest -c polaris/pytest.ini `
Signed-off-by: Lawrence Lawson <lawrence@stellar.org>


polaris/polaris/utils.py
```diff
@@ -135,24 +135,23 @@ def create_stellar_deposit(
         transaction.save()
         raise ValueError(transaction.status_message)
 
-    # if we don't know if the destination account exists
-    if not destination_exists:
```
JakeUrban, line 139: So you're right we don't need to check this because the destination account should always exist.

But that means `destination_exists` doesn't even need to be a parameter, and all invocations of `create_stellar_deposit()` with that parameter should be adjusted.

polaris/polaris/utils.py
```diff
@@ -135,24 +135,23 @@ def create_stellar_deposit(
...
+    except RuntimeError as e:
+        transaction.status = Transaction.STATUS.error
+        transaction.status_message = str(e)
+        transaction.save()
+        logger.error(transaction.status_message)
+        return False
+    if has_trustline(transaction, json_resp):
+        # the account is doesn't trust the asset to be received
+        if transaction.status != Transaction.STATUS.pending_trust:
+            transaction.status = Transaction.STATUS.pending_trust
```
JakeUrban, line 152: Now that I'm seeing the code change, I'm realizing it may be unnecessary to split up the removal of the `pending_trust` logic. Because if we remove the `get_or_create_transaction_destination_account()` invocation AND the `pending_trust` logic in one change, we can just remove this entire section right?

polaris/polaris/utils.py
```diff
@@ -135,24 +135,23 @@ def create_stellar_deposit(
         transaction.save()
         raise ValueError(transaction.status_message)
 
-    # if we don't know if the destination account exists
-    if not destination_exists:
-        try:
-            _, created, pending_trust = get_or_create_transaction_destination_account(
```
JakeUrban, line 141: I just looked at where else `get_or_create_transaction_destination_account()` is called, and its only in `execute_deposits()`, so can we move this function to the `poll_pending_deposits.py` file?

polaris/polaris/utils.py
JakeUrban: And if we're removing the `pending_trust` recovery functionality, we also do it in `submit_stellar_deposit()`, so we'll need to remove it there too.

JakeUrban (changes requested):
Nice work, this accomplishes exactly what I said, but now that I'm looking at it I'm realizing it will be an even simpler change to just remove the `pending_trust` recovery functionality in this PR as well.

polaris/polaris/utils.py
Zagan202: I don't think I understand, remove the entire section try block or     
```
if has_trustline(transaction, json_resp):
        # the account is doesn't trust the asset to be received
        if transaction.status != Transaction.STATUS.pending_trust:
            transaction.status = Transaction.STATUS.pending_trust
```

JakeUrban: If we remove that section you just referenced, do we need the try block?

Zagan202: So in that case we should remove the test `def test_deposit_stellar_no_account(acc1_usd_deposit_transaction_factory):` since it serves no plausible scenario where the account doesnt exist and must be created.

Zagan202: I will have to put some of this logic back in if I want to tell if the given account has a `trustline` and is `claimable_balance_supported`.

polaris/polaris/management/commands/poll_pending_deposits.py
```diff
@@ -83,6 +87,90 @@ def check_for_multisig(transaction):
...
+
+        transaction.status = Transaction.STATUS.pending_trust
+        transaction.save()
+        logger.info(
+            f"Transaction {transaction.id} is now pending_trust of destination account"
+        )
+        account, json_resp = get_account_obj(
+            Keypair.from_public_key(transaction.stellar_account)
+        )
+        return account, True, has_trustline(transaction, json_resp)
```
JakeUrban, line 169: No need to check the JSON response of an account that was just created, it definitely doesn't have a trustline.
```suggestion
        account, _ = get_account_obj(
            Keypair.from_public_key(transaction.stellar_account)
        )
        return account, True, False
```

JakeUrban (approved):
LGTM! One small change I just noticed

polaris/polaris/management/commands/poll_pending_deposits.py
Zagan202: Actually I believe this need to be `True`
Unless I'm misreading this `has_trustlines` returns false if there is a trustline and true if there isnt
This is probable because of the lose association with the `pending_trust` status
```
def has_trustline(transaction, json_resp):
    pending_trust = True
    for balance in json_resp["balances"]:
        if balance.get("asset_type") == "native":
            continue
        asset_code = balance["asset_code"]
        asset_issuer = balance["asset_issuer"]
        if (
            transaction.asset.code == asset_code
            and transaction.asset.issuer == asset_issuer
        ):
            pending_trust = False
            break
    return pending_trust
```

JakeUrban (approved):
You’re right, good catch

JakeUrban:
Oh, one last thing: remove the pending_trust logic from submit_stellar_deposit as well

JakeUrban (approved):

PR #375, opened 2021-01-20
update translation locale id for v1.2

oonid:
update polaris/polaris/locale/id/
update example/server/locale/id/


JakeUrban (approved):
Thanks!

PR #403, opened 2021-03-31
Add an explanatory command to tutorial

kjreills:
While setting up an anchor POC, I did not realize at first that I needed to use the `migrate.py` shell, I was just using the regular python shell and receiving errors. I think this could help explain that the user needs to perform this specific action to get into the shell with the management context loaded.


JakeUrban:
Ah, that would be helpful, thanks 👍

PR #415, opened 2021-04-15
SEP-24: make amount input symbol padding variable

### LINKED ISSUE #406: SEP24 amount symbol overflow
Currently, if `asset.symbol` has many characters, the symbol overflows the amount input field:

image

A workaround is to dynamically set the input padding via JS:
```
const icon = document.querySelector('#id_amount').previousElementSibling;
const iconWidthStr = getComputedStyle(icon).width;
const iconWidthFloat = parseFloat(iconWidthStr.replace(/px/,""));
const iconPaddingLeftStr = getComputedStyle(icon).paddingLeft;
const iconPaddingLeftFloat = parseFloat(iconPaddingLeftStr.replace(/px/,""));
const amountElem = document.getElementById('id_amount');
amountElem.style.paddingLeft = (iconWidthFloat + iconPaddingLeftFloat) + 'px';
```

Anchors might wanna display the asset code instead of the symbol to make it more clear to the user what's the currency being deposited/withdrawn, but it would require a workaround to prevent that overflow.
Polaris could support many characters in the symbol by default.

JakeUrban:
resolves #406 

Makes the `.amount-input` element's `paddingLeft` attribute depend on the width of `.icon`. This PR also adds a space between the symbol and amount in the fee table.

cc @yuriescl

### DIFF
--- a/polaris/polaris/templates/polaris/base.html
+++ b/polaris/polaris/templates/polaris/base.html
@@ -142,6 +142,13 @@
         })
       }
 
+      // Amount symbol padding
+      const amountInputElem = document.querySelector('.amount-input');
+      const icon = amountInputElem.previousElementSibling;
+      const iconWidthStr = getComputedStyle(icon).width;
+      const iconWidthFloat = parseFloat(iconWidthStr.replace(/px/, ''));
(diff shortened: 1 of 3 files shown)


yuriescl (commented):
thanks!

Issue #427, opened 2021-05-26
amount_fee not set during SEP-24 flow


yuriescl:
Is there a reason Polaris doesn't set `Transaction.amount_fee` automatically during SEP-24 flow (
https://github.com/stellar/django-polaris/blob/master/polaris/polaris/sep24/deposit.py#L127), but does set it in `poll_pending_deposits` (https://github.com/stellar/django-polaris/blob/master/polaris/polaris/management/commands/poll_pending_deposits.py#L164)?

SEP-24 Fee field will display `None` until the deposit is executed:

![image](https://user-images.githubusercontent.com/26092447/119704368-f2213380-be2d-11eb-9890-beaae50da426.png)

But it does display the fee when filling the amount:
![image](https://user-images.githubusercontent.com/26092447/119701912-41199980-be2b-11eb-8ef0-fc1e7bcd3deb.png)

JakeUrban:
The original idea was to not display a fee until the funds were delivered to the user. However I think we match the behavior of the `amount_in` field, which has a "To Be Received" label prior to the anchor receiving funds. We can do the same for "Amount Deposited" --> "Amount To Be Deposited".

yuriescl:
> We can do the same for "Amount Deposited" --> "Amount To Be Deposited"

I agree

About `amount_fee`, I think it could be initialized in the same places as `amount_in`. The anchor usually has to handle `amount_in` and `amount_fee` anyway in the rails code, so having an `amount_fee` there already wouldn't make much difference.

JakeUrban:
After thinking about this some more I'm realizing that the SEPs don't clarify whether or not `amount_fee` and `amount_out` can or should be non-null before the transaction is completed. Right now, Polaris doesn't populate `amount_fee` or `amount_out` until the transaction is complete.

It is possible that a client application checks `amount_out` to determine if funds have been delivered, but I'm not sure how common that practice is. `status` is the source of truth for the state of the transaction, and clients _should_ be using this attribute instead of `amount_out`.

I think we should clarify in the SEPs that `amount_in`, `amount_fee`, and `amount_out` can (and probably should) be assigned a value as soon as the values can be calculated. In the same change we should reiterate that `status` is the attribute that communicates whether or not funds have been received by the anchor and delivered to the user.

I'll make the change in Polaris now but won't release it until the clarifications to the SEPs have been merged.

JakeUrban:
[...]

PR #429, opened 2021-05-27
SEP-6 & 24: save amount_fee and amount_out asap, adjust SEP-24 more_info labels

JakeUrban:
This PR changes the `example` application's behavior by calculating `Transaction.amount_fee` and `Transaction.amount_out` as soon as `Transaction.amount_in` is provided by the user.

The one change made to the Polaris library is that the SEP-24 `more_info.html` template will now use the labels `Amount To Be Deposited` and `Amount To Be Withdrawn` if `Transaction.status` is not `completed` AND a value is assigned to `Transaction.amount_out`.


JakeUrban:
@yuriescl

relates to #427

PR #432, opened 2021-06-01
SEP-24: Detect `TransactionForm.type` attribute and use in `GET /fee` requests

### LINKED ISSUE #427: amount_fee not set during SEP-24 flow

JakeUrban:
partially resolves #427

Updates the JavaScript code in `base.html` to detect a type form attribute using the `id_type` HTML ID and uses it in calls to `GET /fee`. Also updates the reference server to use the functionality and updates `TransactionForm` documentation.

cc @yuriescl

### DIFF
--- a/example/server/forms.py
+++ b/example/server/forms.py
@@ -26,6 +26,10 @@ class KYCForm(forms.Form):
 class WithdrawForm(TransactionForm):
     """This form accepts the amount to withdraw from the user."""
 
+    type = forms.ChoiceField(
+        choices=[(None, "--"), ("bank_account", "Bank Account")],
+        label="Transaction Type",
+    )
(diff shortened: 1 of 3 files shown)


yuriescl:
Thanks!

Issue #445, opened 2021-07-07
watch_transactions crashing due to Postgres sorting of NULL fields


ronaldoy:
watch_transactions keeps crashing with 503 or 504 errors when trying to get the last_completed_transaction and some transactions have completed_at = NULL. It happens because Postgres defaults to NULLS FIRST.

JakeUrban:
As described in #446 , Polaris always assigns `completed_at` when the `completed` status is also assigned.

PR #446, opened 2021-07-07
Fix watch_transactions crashes with Postgres (#445)

ronaldoy:


JakeUrban:
If a transaction's status is `completed`, its `completed_at` timestamp should be non-null, so this statement should have no effect.

I've checked every place where transaction statuses are assigned the `completed` status, and only found one place where its possible for `completed_at` to be null, but its an error case: [source](https://github.com/stellar/django-polaris/blob/master/polaris/polaris/management/commands/execute_outgoing_transactions.py#L160).

If you can find a place where `completed_at` should be assigned but it is not, then we can fix that bug.

PR #447, opened 2021-07-13
SEP-24 added support to form hidden fields

yuriescl:
Some Anchors might wanna use form hidden fields to identify which form is being submitted to `form_for_transaction` integration function.
For example, if there are two forms, the user fills the first one and submits it, then gets redirected to the next form, but then presses the browser back button, then submits the first form again, the Anchor can know the user is re-submitting the first form by, for example, looking at the value of a certain hidden field:

```
class WithdrawalForm0(forms.Form):
    step = forms.CharField(
        widget=forms.HiddenInput(attrs={'value': '0'}),
        required=True
    )
   ...

class WithdrawalForm1(forms.Form):
    step = forms.CharField(
        widget=forms.HiddenInput(attrs={'value': '1'}),
        required=True
    )
   ...
```

### DIFF
--- a/polaris/polaris/templates/polaris/deposit.html
+++ b/polaris/polaris/templates/polaris/deposit.html
@@ -17,7 +17,11 @@
     <form method="post" action="{{ post_url }}">
       {% csrf_token %}
 
-      {% for field in form %}
+      {% for hidden in form.hidden_fields %}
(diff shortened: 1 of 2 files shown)


JakeUrban (approved):
This makes sense to add, thanks.

Issue #449, opened 2021-07-20
Not able to Deposit and Withdraw Money by using Custom Stellar Anchor in Demo Wallet


petluru:
I built a Stellar Anchor Service under home domain "poc.utilli.co".

You can find the Stellar Toml file for that service with this link. https://poc.utilli.co/.well-known/stellar.toml

Problem that I am facing:

I am unable to make deposit and withdrawl transactions with my anchor service on "demo-wallet". I was able to see the Asset that I Anchored and was able to create trustline to the login account.

But, during the SEP-24 Deposit and Withdrawl the Pop-up for user interactivity is not showing the UI for entering the amount and KYC details.

On debugging using the console, the anchor is hosted in home domain which is "https", whereas the url I am getting while doing the SEP-24 deposit is not "https".

While checking in the Stellar Anchor Tests service

URL: https://anchor-tests.stellar.org

I am getting following error in sep-24:

![SEP-24 Test Anchor Issue](https://user-images.githubusercontent.com/36133568/126336144-7e9dbcf4-a270-4cc1-a06a-9f8561f93a72.png)

I have enabled CORS in the Django Settings.py file.

Also when trying to see the API that is called when SEP-24 Deposit is done I am getting following URL

http://poc.utilli.co/sep24/transactions/deposit/webapp?asset_code=XXXXX&transaction_id=XXXXXXX

For confidentiality I am hiding the asset_code and transaction_id. But, the point of my concern is API is called in http not in https.

Anyone, could you please suggest me what is the mistake that I am doing.

I followed all the steps metioned in the django-polaris readthedocs website. https://django-polaris.readthedocs.io/

JakeUrban:
Hi @petluru, you're returning an 'http' URL in response to the POST /deposit request:

![Screen Shot 2021-07-20 at 7 51 05 AM](https://user-images.githubusercontent.com/10968980/126345489-1810ed6f-689d-45bd-8aec-c58dffed23b3.png)

petluru:
Hi Yuri, I am using AWS Elastic Load Balancer based certificate for this. Currently the home domain of the anchor is https. So, only proxy needs to be used along with nginx ?
Can I use Apache also ?

yuriescl:
> Hi Yuri, I am using AWS Elastic Load Balancer based certificate for this. Currently the home domain of the anchor is https. So, only proxy needs to be used along with nginx ?
> Can I use Apache also ?

[...]

PR #450, opened 2021-07-20
Added Nginx example configuration

yuriescl:
Anchors might need a reference configuration for Nginx. It took me some time to get it working properly, specially the HTTPS and proxy.

### DIFF
--- a/example/nginx-letsencrypt.conf
+++ b/example/nginx-letsencrypt.conf
@@ -1,0 +1,40 @@
+# Example configuration for Django Polaris using Nginx as reverse proxy with HTTPS enabled.
+# https://github.com/stellar/django-polaris
+
+# This configuration assumes:
+# 1) Polaris is running on localhost:8000
+# 2) Let's Encrypt Certbot was used for auto generating the SSL config lines
+
+server {
+    server_name stellar-anchor.mydomain.com;
+
+    location / {
+        proxy_pass http://localhost:8000;
+        proxy_set_header Host $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_redirect http://localhost:8000 https://stellar-anchor.mydomain.com;
+    }
+
+    listen [::]:443 ssl ipv6only=on; # managed by Let's Encrypt Certbot
+    listen 443 ssl; # managed by Let's Encrypt Certbot
+    ssl_certificate ...; # managed by Let's Encrypt Certbot
+    ssl_certificate_key ...; # managed by Let's Encrypt Certbot
+    include /etc/letsencrypt/options-ssl-nginx.conf; # managed by Let's Encrypt Certbot
+    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem; # managed by Let's Encrypt Certbot
+
+}
+
+server {
+    if ($host = stellar-anchor.mydomain.com) {
+        return 301 https://$host$request_uri;
+    } # managed by Let's Encrypt Certbot
+
+    listen 80 default_server;
+    listen [::]:80 default_server;
+
+    server_name stellar-anchor.mydomain.com;
+    return 404; # managed by Let's Encrypt Certbot
+}
+
+
--- a/example/nginx-letsencrypt.conf
+++ b/example/nginx-letsencrypt.conf
@@ -13,7 +13,6 @@ server {
         proxy_set_header Host $host;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header X-Forwarded-Proto $scheme;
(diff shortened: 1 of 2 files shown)


JakeUrban (commented):
Just to confirm my understanding, the following env variables would be used with this configuration:

- LOCAL_MODE: false
- SECURE_SSL_REDIRECT: true
- SECURE_PROXY_SSL_HEADER: ("HTTP_X_FORWARDED_PROTO", "https")

Is your `HOST_URL` `http://localhost:8000` or `stellar-anchor.mydomain.com`?

yuriescl:
> Just to confirm my understanding, the following env variables would be used with this configuration:
> 
>     * LOCAL_MODE: false
> 
>     * SECURE_SSL_REDIRECT: true
> 
>     * SECURE_PROXY_SSL_HEADER: ("HTTP_X_FORWARDED_PROTO", "https")
> 
> 
> Is your `HOST_URL` `http://localhost:8000` or `stellar-anchor.mydomain.com`?

This is what I'm using:

```
POLARIS_LOCAL_MODE = False
SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
SECURE_SSL_REDIRECT = False  # Nginx does the redirection
```

yuriescl:
> Is your HOST_URL http://localhost:8000 or stellar-anchor.mydomain.com?
`POLARIS_HOST_URL='https://stellar-anchor.mydomain.com'`

yuriescl:
I have had issues with SEP-24 because Polaris was generating the interactive URL with `http://`.
I had to set this line in the Nginx config to fix that:
`proxy_redirect http://localhost:8000 https://stellar-anchor.mydomain.com`
Otherwise Polaris would give something like:
```
{
  "type": "interactive_customer_info_needed",
  "url" : "http://stellar-anchor.mydomain.com.....",
  "id": "82fhs729f63dh0v4"
}
```

yuriescl:
https://github.com/stellar/django-polaris/issues/449 is related to this. It's a reverse proxy configuration issue, that's why I think we should have a reference Nginx config file on the repo, so Anchors can spot which configuration they might be missing.

JakeUrban:
> I have had issues with SEP-24 because Polaris was generating the interactive URL with `http://`.
> I had to set this line in the Nginx config to fix that:
> `proxy_redirect http://localhost:8000 https://stellar-anchor.mydomain.com`
> Otherwise Polaris would give something like:
> 
> ```
> {
>   "type": "interactive_customer_info_needed",
>   "url" : "http://stellar-anchor.mydomain.com.....",
>   "id": "82fhs729f63dh0v4"
> }
> ```

Nginx's `proxy_redirect` seems to only concern itself with the `Location` and `Refresh` response headers, not an attribute within a JSON response body. I would imagine that `SECURE_PROXY_SSL_HEADER` being set informs Django to construct HTTPS URLs if the original request used HTTPS, are you sure the `url` JSON attribute uses http without the `proxy_redirect` directive?

JakeUrban:
Looking at the [Django source code](https://github.com/django/django/blob/8bcb00858e0ddec79cc96669c238d29c30d7effb/django/http/request.py#L231), it seems `SECURE_PROXY_SSL_HEADER` is all you need.

yuriescl:
> Looking at the [Django source code](https://github.com/django/django/blob/8bcb00858e0ddec79cc96669c238d29c30d7effb/django/http/request.py#L231), it seems `SECURE_PROXY_SSL_HEADER` is all you need.

Just tested, that is correct. My reverse proxy config is set like this and it's correctly returning `https` in the JSON url:
`settings.py`:
```
SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
SECURE_SSL_REDIRECT = False
```
`nginx.conf`:
```
    location / {
        proxy_pass http://localhost:8000;
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header X-Forwarded-Proto $scheme;
    }
```
For some reason that I don't remember I had to add `proxy_redirect http://localhost:8006 https://stellar-anchor.mydomain.com` to Nginx. It's probably not required.

example/nginx-letsencrypt.conf
```diff
@@ -0,0 +1,39 @@
+# Example configuration for Django Polaris using Nginx as reverse proxy with HTTPS enabled.
```
JakeUrban, line 1: Can you add a comment specifying the linux distribution used? The only reason this is necessary is due to the file-system paths generated by Lets Encrypt, which may not be the correct paths on other linux distributions.

Lets also mention that this is not used in the deployment of this example application, but is purely for reference.

JakeUrban (commented):
A couple more requests to add some comments, but after they're made I'll approve 👍

example/nginx-letsencrypt.conf
yuriescl: Done, thanks

JakeUrban (approved):

PR #468, opened 2021-08-08
paging_token blank=True

yuriescl:
Sometimes, in the Admin panel, we (Anchor) have to manually update a `Transaction`, but we get an error saying `paging_token` is required (but it doesn't contain a value yet). So we always have to put some arbitrary value there to be able to save.
This pull request makes the `paging_token` optional in forms.


yuriescl:
@JakeUrban This pull request is incomplete as it doesn't contain the migrations (and maybe other necessary changes). It's more like a proposal/idea.

PR #487, opened 2021-08-18
Show asterisk for required fields

yuriescl:
This is a minor UI improvement. It adds an asterisk after the field label if the field is required. It helps the user to know which fields are required in a given form.

I'm not very familiar with SCSS and compiling it to CSS, so you'd need to add the styling for the `required` class.
I suggest something similar to this:
```
.field label span.required {
    color: red;
    font-weight: bold;
}
```

### DIFF
--- a/polaris/polaris/templates/polaris/deposit.html
+++ b/polaris/polaris/templates/polaris/deposit.html
@@ -24,7 +24,12 @@
       {% for field in form.visible_fields %}
         <div class="field">
           <label>
-            <div class="label">{{ field.label }}</div>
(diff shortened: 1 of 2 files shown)


JakeUrban:
Thanks, this is probably something the UI should have. Although I think Django will highlight the field in a red outline if the input is selected and then deselected without an input. Still, we could indicate the required fields before the user tries to submit.

JakeUrban:
I may recreate this PR on my origin so I can push to it, but I'll use this PR for tracking

PR #504, opened 2021-08-31
Configure Polaris Reference Server to run locally without Docker and integrate with PyCharm

lijamie98:


docs/contribute/development.rst
```diff
@@ -0,0 +1,65 @@
+# How to set up Polaris development environment
+
+If you would like to contribute to Polaris, this document will guide you to set up your development environment on
+Ubuntu systems.
+
+## Ubuntu 20.04LTS
```
JakeUrban, line 6: Were you planning to add different instructions for Ubuntu 18? Or is it the same? If its the same flow I would remove this header.

JakeUrban (commented):
This looks great so far. Lets update the docs structure to include this document and reference it in the "Contributing" section of `index.rst`.

Also, we should probably go a bit futher than just installing the correct python version. Can we add instructions for installing pipenv, pulling the django-polaris repo, and any other setup needed to get the reference server running via docker-compose?

docs/contribute/development.rst
lijamie98: Yes, I will add instructions for installing pip, pipenv, pulling django-polaris to the document. I will find a way to add docker-compose instruction.

lijamie98: I will remove the header. I don't think it makes sense to use Ubuntu 18 when we have 20.04LTS

PR #506, opened 2021-08-31
Added django-polaris package path for local development without Docker. 

lijamie98:


example/Pipfile
```diff
@@ -9,6 +9,11 @@ watchdog = "<0.10"
 [packages]
 gunicorn = "*"
 psycopg2-binary = ">=2.9, <3.0"
+
+# Uncomment the following line if you would like to run the dev without docker
```
JakeUrban, line 13:
```suggestion
# Uncomment the following line if you would like to run the example without docker
```

example/Pipfile
```diff
@@ -9,6 +9,11 @@ watchdog = "<0.10"
...
+# django-polaris = {path = "..", editable = true}
+
+# Comment the following line if you would like to run the dev without docker
```
JakeUrban, line 16:
```suggestion
# Comment the following line if you would like to run the example without docker
```

example/Pipfile
```diff
@@ -9,6 +9,11 @@ watchdog = "<0.10"
 [packages]
 gunicorn = "*"
 psycopg2-binary = ">=2.9, <3.0"
+
+# Uncomment the following line if you would like to run the dev without docker
+# django-polaris = {path = "..", editable = true}
```
JakeUrban, line 14: Is `..` correct? Is this path relative to the current working directory? For example, if I run from the `example` directory does the path need to be different than if I run from the `django-polaris` directory?

JakeUrban (approved):
A couple suggestions and one question. But if they're all resolved then it LGTM

example/Pipfile
lijamie98: Will change

example/Pipfile
lijamie98: Will change

example/Pipfile
lijamie98: Cannot find a way to inject a path with something like $HOME/django-polaris. 

The `pipenv install --dev` must be run from the example folder. This issue can be taken care of in the document.

PR #523, opened 2021-09-11
SEP-24: fix amount field and fee table localization

JakeUrban:
`polaris.integrations.forms.TransactionForm` was improperly initializing the amount field, resulting in excessive character restrictions on the decimal field. This wasn't a problem for english anchors, but for anchors in other locales, users would not have been able to specify decimal inputs properly.

This PR also improves localization in various ways - such as number rendering in `more_info.html` and enabling client-side javascript to parse numbers formatted for other locales to calculate fees.

cc @ronaldoy


lijamie98 (approved):
I am not a JS expert. But LGTM.

PR #532, opened 2021-09-21
SEP-38: RFQ API Implementation

lijamie98:


example/server/integrations.py
```diff
@@ -970,3 +988,161 @@ def info_integration(request: Request, asset: Asset, lang: str):
...
+class MySEP38AnchorIntegration(SEP38AnchorIntegration):
+    def get_prices(
+        self,
+        sell_asset: str,
+        sell_amount: str,
+        sell_delivery_method: str = None,
+        buy_delivery_method: str = None,
+        country_code: str = None,
+    ) -> List[GetPricesResponse]:
+        exchange_pairs = list_exchange_pairs(buy_asset=sell_asset)
```
JakeUrban, line 1024: I can see possibility for confusion here with `sell_asset/buy_asset`. Why don't we try to remove ambiguity by using the terms `client` and `anchor`? Ex. `list_exchange_pairs(anchor_buy_asset=client_sell_asset)`.

example/server/integrations.py
```diff
@@ -970,3 +988,161 @@ def info_integration(request: Request, asset: Asset, lang: str):
...
+
+        prices = []
+        for exchange_pair in exchange_pairs:
+            if is_stellar_asset(sell_asset):
```
JakeUrban, line 1028: This call can be made outside the `for` loop, since `sell_asset` is defined in the function parameters.

example/server/integrations.py
```diff
@@ -970,3 +988,161 @@ def info_integration(request: Request, asset: Asset, lang: str):
             }
         },
     }
+
+
+def check_delivery_method(
```
JakeUrban, line 993: can we rename this `deliver_method_supported()`?

example/server/integrations.py
```diff
@@ -970,3 +988,161 @@ def info_integration(request: Request, asset: Asset, lang: str):
...
+    ) -> List[GetPricesResponse]:
+        exchange_pairs = list_exchange_pairs(buy_asset=sell_asset)
+
+        prices = []
+        for exchange_pair in exchange_pairs:
+            if is_stellar_asset(sell_asset):
+                # We are selling a stellar asset. Buying asset is offchain.
+                if not check_delivery_method(
+                    asset=exchange_pair.sell_asset,
+                    list_delivery_methods_func=get_buy_delivery_methods,
```
JakeUrban, line 1032: I'm not a fan of passing functions to other functions in Python, can we write these utility functions in a way that doesn't require this?

example/server/integrations.py
```diff
@@ -970,3 +988,161 @@ def info_integration(request: Request, asset: Asset, lang: str):
...
+        sell_delivery_method: str = None,
+        buy_delivery_method: str = None,
+        country_code: str = None,
+    ) -> List[GetPricesResponse]:
+        exchange_pairs = list_exchange_pairs(buy_asset=sell_asset)
+
+        prices = []
+        for exchange_pair in exchange_pairs:
+            if is_stellar_asset(sell_asset):
+                # We are selling a stellar asset. Buying asset is offchain.
```
JakeUrban, line 1029: In the API, `sell_asset` is defined as the asset the client wants to sell. So if `is_stellar_asset(is_sell_asset)` is `True`, that means the anchor is buying a stellar asset, and selling an off-chain asset.

example/server/integrations.py
```diff
@@ -970,3 +988,161 @@ def info_integration(request: Request, asset: Asset, lang: str):
...
+) -> bool:
+    if delivery_method:
+        bdms = list_delivery_methods_func(get_offchain_asset(asset))
+        if not any(bdm.name == delivery_method for bdm in bdms):
+            # This method is not supported
+            return False
+    return True
+
+
+def check_country_code(asset: str, country_code: str) -> bool:
```
JakeUrban, line 1006: Rename to `country_code_supported()`?

example/server/integrations.py
JakeUrban: Maybe instead of the `asset` parameter we can have `sell_asset` and `buy_asset` so the `check_delivery_method()` function knows which `get_*_delivery_methods()` function to call.

polaris/polaris/utils.py
```diff
@@ -349,3 +350,15 @@ def validate_patch_request_fields(fields: Dict, transaction: Transaction):
...
+def to_decimals(float_value: float, precision: int) -> str:
+    """
+    Converting from float to string with the decimal precision.
+    """
+    return str(
+        decimal.Decimal(str(float_value)).quantize(
+            decimal.Decimal(".{}1".format("0" * (precision - 1))),
+            rounding=decimal.ROUND_05UP,
+        )
+    )
```
JakeUrban, line 364: I'm not sure we should provide this as a utility function. It has the assumption that `float_value` is less than 0 for example, so it seems pretty specific to a particular use case.

I also see most of the calls to this function casting their `float_value` parameter, only to have it cast back to a string here.

Have you seen how Polaris handles this logic in other parts of the code?

```
                transaction.amount_out = round(
                    transaction.amount_in - transaction.amount_fee,
                    transaction.asset.significant_decimals,
                )
```

Where `amount_in`, `amount_out`, and `amount_fee` are `Decimal` instances. I would avoid working with floats in general.

example/server/integrations.py
```diff
@@ -970,3 +988,161 @@ def info_integration(request: Request, asset: Asset, lang: str):
...
+        sell_amount: str = None,
+        buy_amount: str = None,
+        sell_delivery_method: str = None,
+        buy_delivery_method: str = None,
+        country_code: str = None,
+    ) -> GetPriceResponse:
+        exchange_pairs = list_exchange_pairs(buy_asset=sell_asset, sell_asset=buy_asset)
+        # Must have exactly only one price found.
+        if len(exchange_pairs) == 0:
+            raise ValidationError(f"No price is found:  {sell_asset}->{buy_asset}")
```
JakeUrban, line 1077: This is another new pattern; raising Django's `ValidationError` exception vs. a builtin exception like `ValueError`. Also, can't Polaris validate this before calling `get_price()`? If so, then we can validate it in the endpoint code and mention in the docstring that this validation has taken place already.

example/server/urls.py
```diff
@@ -13,13 +13,15 @@
...
+from django.urls import path, include
+
 import polaris.urls
+from polaris.sep38 import manager_api
 from .views import all_fields_form_view, confirm_email, skip_confirm_email, log_callback
 
-
 urlpatterns = [
+    path("admin/", admin.site.urls),
+    path("anchor/sep38/", include(manager_api)),
```
JakeUrban, line 24: Lets remove both of these URL paths from the reference server, they're not needed.

polaris/polaris/integrations/quote.py
```diff
@@ -0,0 +1,57 @@
...
+class SEP38AnchorIntegration:
+    """
+    The container class for SEP38 integrations
+
+    The list of the offline assets supported by the anchor should be returned.
+
+    The returned dictionary should contains the following attributes specified at:
+    https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#get-info
+
+    If there is no Offchain assets, an empty list is expected.
```
JakeUrban, line 27: This seems like a docstring for a function, not the entire class

polaris/polaris/integrations/rails.py
```diff
@@ -166,5 +166,13 @@ def poll_pending_deposits(
         """
         raise NotImplementedError()
 
+    def get_indicative_price(self, buy_asset: str, sell_asset: str) -> str:
+        pass
+
+    def get_firm_price(
+        self, buy_asset: str, sell_asset: str, amount: str, expire_after: str
+    ) -> str:
+        pass
```
JakeUrban, line 175: I don't think these should be integration functions on `RailsIntegration` class. I don't think we need to provide an interface for Polaris to request these prices from developers in general, all we need are the functions defined on `SEP38AnchorIntegration`.

polaris/polaris/migrations/0012_buydeliverymethod_exchangepair_offchainasset_quote_selldeliverymethod.py
```diff
@@ -0,0 +1,133 @@
...
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("schema", models.TextField()),
+                ("identifier", models.TextField()),
+                ("significant_decimals", models.IntegerField(default=2)),
+                ("country_codes", models.TextField(blank=True)),
```
JakeUrban, line 47: we need `null=True` on this as well.

polaris/polaris/models.py
```diff
@@ -672,3 +669,101 @@ def channel_account(self):
...
+
+    sell_delivery_method = models.TextField(null=False, blank=True)
+    """
+    One of the name values specified by the sell_delivery_methods array.
+    """
+
+    buy_delivery_method = models.TextField(null=False, blank=True)
+    """
+    One of the name values specified by the buy_delivery_methods array.
+    """
```
JakeUrban, line 719: Why wouldn't these be foreign keys to the models?

polaris/polaris/sep38/info.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # Populate offchain assets
+    offchain_assets = list_offchain_assets()
+
+    if offchain_assets is None:
+        return render_error_response(
+            "The anchor returned None as a list of offchain-asset. An empty list is expected.",
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            content_type="text/html",
+        )
```
JakeUrban, line 34: `list_offchain_assets()` is a function defined by Polaris that always returns a list, why are we checking if its `None` here?

polaris/polaris/sep38/manager_api.py
```diff
@@ -0,0 +1,232 @@
+from django.forms import model_to_dict
```
JakeUrban, line 1: I know this wasn't a trivial piece of code to write but there isn't a need for it, and allows external clients to change the anchor's data model, which is a security concern. Lets remove this file.

polaris/polaris/sep38/prices.py
```diff
@@ -0,0 +1,92 @@
...
+    _validate_asset(r.get("sell_asset"))
+    _validate_asset(r.get("buy_asset"))
+    if not ((r.get("sell_amount") is not None) ^ (r.get("buy_amount") is not None)):
+        raise ValueError(
+            "Exactly one of `buy_amount` and `set_amount` fields must be set."
+        )
+    if r.get("sell_amount"):
+        r["sell_amount"] = int(r.get("sell_amount"))
+    if r.get("buy_amount"):
+        r["buy_amount"] = int(r.get("buy_amount"))
```
JakeUrban, line 72: Why are we casting these amounts as integers? We need to allow amounts up to the number of decimals supported for these assets.

polaris/polaris/sep38/prices.py
```diff
@@ -0,0 +1,92 @@
...
+@api_view(["GET"])
+@renderer_classes([JSONRenderer])
+@parser_classes([JSONParser])
+@check_authentication()
+def get_price(request: Request) -> Response:
+    try:
+        request_data = _validate_price_request(request.data)
+        qp = rqi.get_price(**request_data)
+        return Response(qp.__dict__)
+    except ValidationError as ex:
```
JakeUrban, line 86: You're catching `ValidationError` but raising `ValueError` in `_validate_price_request()`.

polaris/polaris/sep38/prices.py
```diff
@@ -0,0 +1,92 @@
...
+    results = []
+    for it in indicative_prices:
+        results.append(it.__dict__)
+    return Response(results)
+
+
+def _validate_price_request(r: dict) -> dict:
+    _validate_asset(r.get("sell_asset"))
+    _validate_asset(r.get("buy_asset"))
+    if not ((r.get("sell_amount") is not None) ^ (r.get("buy_amount") is not None)):
```
JakeUrban, line 65: TIL: XOR in python

polaris/polaris/sep38/quote.py
```diff
@@ -0,0 +1,119 @@
...
+from polaris.sep38.utils import (
+    get_significant_decimals,
+    get_quote_by_id,
+    get_exchange_pair,
+)
+from polaris.utils import render_error_response, to_decimals
+
+
+def validate_quote_request(r: dict) -> dict:
+    r = r.copy()
```
JakeUrban, line 23: I don't see a reason to copy this dict.

polaris/polaris/sep38/quote.py
```diff
@@ -0,0 +1,119 @@
...
+    @check_authentication()
+    def get(_: Request, quote_id: str) -> Response:
+        try:
+            quote = get_quote_by_id(quote_id)
+            return Response(model_to_dict(quote, fields=quote_response_fields))
+        except Exception as ex:
+            return render_error_response(
+                description="Error getting the quote. cause={}".format(ex.__str__()),
+                status_code=status.HTTP_400_BAD_REQUEST,
+            )
```
JakeUrban, line 67: Lets not assume all exceptions are the client's fault. Can we catch the exceptions we know will be raised as a result of a bad client request?

polaris/polaris/sep38/quote.py
```diff
@@ -0,0 +1,119 @@
...
+
+        quote = Quote()
+        quote.sell_asset = r.get("sell_asset")
+        quote.buy_asset = r.get("buy_asset")
+        quote.sell_amount = r.get("sell_amount")
+        quote.buy_amount = r.get("buy_amount")
+        quote.requested_expire_after = r.get("expire_after")
+        quote.sell_delivery_method = r.get("sell_delivery_method")
+        quote.buy_delivery_method = r.get("buy_delivery_method")
+        quote.country_code = r.get("country_code")
```
JakeUrban, line 90: You can initialize the `Quote` object like so:
```
quote = Quote(
        sell_asset=r.get("sell_asset")
        buy_asset=r.get("buy_asset")
        sell_amount=r.get("sell_amount")
        buy_amount=r.get("buy_amount")
        requested_expire_after=r.get("expire_after")
        sell_delivery_method=r.get("sell_delivery_method")
        buy_delivery_method=r.get("buy_delivery_method")
        country_code=r.get("country_code")
)
```

In fact, if you changed the `expire_after` parameter to be `requested_expire_after` in `validate_quote_request()` you could initialize it like so:

```
quote = Quote(**r)
```

polaris/polaris/sep38/quote.py
```diff
@@ -0,0 +1,119 @@
...
+                    float(quote.buy_amount) * float(quote.price), sell_decimals
+                )
+        except Exception as ex:
+            return render_error_response(
+                description="Error executing the quote. cause={}".format(ex.__str__()),
+                status_code=status.HTTP_400_BAD_REQUEST,
+            )
+
+        Quote.save(quote)
+        return Response(model_to_dict(quote, fields=quote_response_fields))
```
JakeUrban, line 119: Instead of using `model_to_dict()` and similar functions, we should write serializer classes from `rest_framework`. Take a look at our other serializer classes for reference.

polaris/polaris/integrations/quote.py
```diff
@@ -0,0 +1,57 @@
...
+
+    """
+
+    def get_prices(
```
JakeUrban, line 31: We need to pass the `SEP10Token` and `Request` object to every integration function thats called within a SEP-10 authenticated request.

polaris/polaris/sep38/utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def list_offchain_assets() -> List[Dict]:
+    """
+    Gets the list of offchain assets.
+    """
+    result = []
+    for asset in OffChainAsset.objects.all():
+        asset_dict = dict()
+        asset_dict["asset"] = "{}:{}".format(asset.schema, asset.identifier)
```
JakeUrban, line 26: It may make sense to add a property to `OffChainAsset` for this.

JakeUrban (changes requested):
This is a great start, but there are a fair amount of changes I've request. Some are small but some are larger that may require some discussion around design.

JakeUrban (approved):
Approving so we can merge, but we'll still be making changes to the branch.

PR #551, opened 2021-11-02
Add Quote Support for Reference Server

### LINKED ISSUE #476: Implement SEP-38: Anchor RFQ API
Add support for SEP-38 and integrate the functionality into SEP-6 and SEP-31 as defined by the protocols:
- [SEP-38 specification](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md)
- [SEP-31 support for SEP-38](https://github.com/stellar/stellar-protocol/pull/932)
- [SEP-6 support for SEP-38](https://github.com/stellar/stellar-protocol/pull/973)

## Discussion of Changes

The changes below attempt to be exhaustive but are likely not.

### Data Model

There are three approaches we could take:
- Implement as much as possible
- Leave all modeling to the anchor
- Something in between

Implementing as much as possible allows Polaris to handle as much of the quote-related functionality as possible, which minimizes the code the anchor needs to write to have a functioning SEP-38 integration, but forces the anchor to use an existing data model that may be incomplete for their use case. Leaving all modeling to the anchor maximizes the amount of flexibility when designing their own model, but requires them to write more code.

Attempting to add a self-sufficient data model in Polaris would involve not only adding a `Quote` model, but also `OffChainAsset`, `SellDeliveryMethod`, and `BuyDeliveryMethod` models. This would allow us to implement the `GET /info` without an integration function. However, I don't believe this is worth it since all endpoints other than `GET /quote` will require integration functions to source exchange prices.

We can strike a balance and only provide a single `Quote` model:

- `Quote` model
     - `id` - uuid
     - `type` - string (`firm`, or `indicative`)
     - `buy_asset` - string ([Asset Identification Format](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#asset-identification-format))
     - `sell_asset` - string ([Asset Identification Format](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0038.md#asset-identification-format))
     - `price` - nullable(decimal)
     - `sell_amount` - nullable(decimal)
     - `buy_amount` - nullable(decimal)
         - The 3 columns above must be nullable because exchange price for indicative quotes is determined at time of delivery
     - `expires_at` - nullable(UTC datetime)
         - null for indicative quotes

This model will allow us to respond to `GET /quote` requests without an integration function, but the remaining endpoints will require one. Anchors will be expected to create the data model needed for all off-chain asset information.

Changes to existing models:

- `Transaction` model
    - Add `quote` - nullable(foreign key column)
    - Update `kind` - add `deposit-exchange` and `withdrawal-exchange` values

#### SEP-38

All SEP-38 related integration functions will be contained within a `SEP38AnchorIntegration` 

We'll need the following integrations, named after their endpoints:
- `.info()`
- `.prices()`
- `.price()`
- `.quote()`

All integrations will need to be passed the validated request body. `.info()` and `.prices()` will likely need to return the complete response bodies that Polaris will validate before returning to the client. `.price()` will likely only need to return a `Decimal` object and `quote` will likely return a `Quote` model object.

Obviously, we'll need to add all SEp-38 endpoints to the server's url configuration as well.

#### SEP-6

SEP-6 will need to add the following endpoints:
- `GET /deposit-exchange`
- `GET /withdraw-exchange`

I think the pair of these integrations warrant a new integration on the `DepositIntegration` and `WithdrawalIntegration`, `process_sep6_exchange_request()`. It should behave similarly to the original `process_sep6_request()` function.

The functions that validate the response bodies of the `GET /info` endpoint will also need to be adjusted.

The `GET /fee` endpoint will need to be updated to accept the new `kind` parameter values.

The `TransactionSerializer` will need to be updated to return the optional `GET /transaction(s)` response attributes added for transaction using quotes.

#### SEP-31

SEP-31 doesn't have any new endpoints, so it's existing integration functions should work.

The functions that validate the response body of the `GET /info` endpoint will need to be adjusted.

The `SEP31TransactionSerializer` will need to be updated to return the optional `GET /transaction/:id` response attributes added for transaction using quotes.

#### Rails Integrations

CLI-command code that calls the rails integration functions will need to be adjusted to ensure transactions that use indicative quotes have the `Quote.price`, `Quote.buy_amount`, and `Quote.sell_amount` attributes saved after initiating delivery of funds to a user.

Rails integrations & CLI commands will also need to check if the funds were delivered prior to a quote's expiration and handle the case where the quote has expired properly. More thought needs to go into what that entails.

JakeUrban:
resolved #476 

This PR adds support for the reference server to use quotes in SEP-6, 24, and 31.

The reference server will now support 2 off-chain assets, SRT and USD. When initiating a transaction, the client must specify which off-chain asset they want to provide (in the case of a deposit) or receive (in the case of a withdrawal).

The asset collected off-chain will be exchanged for SRT on-chain or vice versa at a rate generated by our ``QuoteIntegration.post_quote()`` implementation. If using SEP-24, users will be asked to "confirm" the exchange rate offered by the anchor.

The only notable change to Polaris integration functions is the introduction of the special `"template_name"` key-value pair that can be included in the dictionary passed back from `.content_for_template()`. If included, Polaris will search for and render the template specified instead of the default templates previously used.


example/server/integrations/deposit.py
```diff
@@ -51,27 +81,73 @@ def content_for_template(
...
+                    "amount_fee": transaction.amount_fee,
+                    "conversion_amount": round(
+                        transaction.amount_in - transaction.amount_fee,
+                        transaction.asset.significant_decimals,
+                    ),
+                    "amount_out": transaction.amount_out,
+                    "amount_in_symbol": "USD",
+                    "amount_fee_symbol": "USD",
+                    "amount_out_symbol": transaction.asset.symbol,
+                    "amount_in_significant_decimals": 2,
```
lijamie98, line 122: Should we sync this up with values in /info?

lijamie98, line 122: May be beneficial to sync up with the SEP-1's `display_decimals` or make them configurable. 

But I think it may be fine since this is for anchor's reference.

example/server/integrations/quotes.py
```diff
@@ -3,6 +3,7 @@
 from decimal import Decimal
 from typing import Union, List, Optional
 
+from django.db.models import Q
```
lijamie98, line 6: unused.

example/server/integrations/withdraw.py
```diff
@@ -6,12 +7,15 @@
 from rest_framework.request import Request
 
 from polaris.integrations import WithdrawalIntegration, calculate_fee
-from polaris.models import Transaction, Asset
+from polaris.models import Transaction, Asset, Quote, OffChainAsset
```
lijamie98, line 10: OffChainAsset unused.

polaris/polaris/management/commands/execute_outgoing_transactions.py
```diff
@@ -81,7 +81,10 @@ def execute_outgoing_transactions():
         sep6_24_qparams = Q(
             protocol__in=[Transaction.PROTOCOL.sep24, Transaction.PROTOCOL.sep6],
             status=Transaction.STATUS.pending_anchor,
-            kind=Transaction.KIND.withdrawal,
+            kind__in=[
```
lijamie98, line 84: Is there a particular reason to use double `_`?

polaris/polaris/management/commands/poll_outgoing_transactions.py
```diff
@@ -65,7 +65,11 @@ def handle(self, *_args, **options):  # pragma: no cover
     @staticmethod
     def poll_outgoing_transactions():
         transactions = Transaction.objects.filter(
-            kind__in=[Transaction.KIND.withdrawal, Transaction.KIND.send],
+            kind__in=[
```
lijamie98, line 68: same as above

polaris/polaris/management/commands/process_pending_deposits.py
```diff
@@ -262,7 +262,10 @@ def get_ready_deposits(cls) -> List[Transaction]:
                     Transaction.STATUS.pending_user_transfer_start,
                     Transaction.STATUS.pending_external,
                 ],
-                kind=Transaction.KIND.deposit,
+                kind__in=[
```
lijamie98, line 265: same as above

polaris/polaris/management/commands/watch_transactions.py
```diff
@@ -121,9 +121,13 @@ async def process_response(cls, response, account):
             return
 
         # Query filters for SEP6 and 24
+
         withdraw_filters = Q(
             status=Transaction.STATUS.pending_user_transfer_start,
-            kind=Transaction.KIND.withdrawal,
+            kind__in=[
```
lijamie98, line 127: same as above

lijamie98 (approved):
Some minor warning fixes and style questions.

example/server/integrations/sep24_kyc.py
```diff
@@ -77,7 +76,7 @@ def track_user_activity(form: forms.Form, transaction: Transaction):
                 else None,
             )
             if server_settings.EMAIL_HOST_USER:
```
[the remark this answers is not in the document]
JakeUrban: Yea I add a comment. This was where we sent email confirmations, but we weren't using a proper email server and the emails often went to spam or failed to send, so its more just for reference

example/server/integrations/deposit.py
JakeUrban: Yes, I forgot to pull the offchain asset model and use `significant_decimals`.

JakeUrban: I need to use `OffChainAsset.significant_decimals`

polaris/polaris/management/commands/execute_outgoing_transactions.py
JakeUrban: Its Django's ORM syntax. Its equivalent to the SQL `SELECT * WHERE kind IN ...`

PR #560, opened 2021-11-11
Update `stellar-sdk`, change dockerfile to use debian base

JakeUrban:


Dockerfile
```diff
@@ -1,12 +1,12 @@
...
-FROM python:3.7-alpine
+FROM python:3.7
 
 # Set environment variables
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
 
 # Install system dependencies
-RUN apk update && apk add build-base postgresql-dev postgresql-client libffi-dev python3-dev cargo gettext-dev curl
+RUN apt-get update && apt-get install -y build-essential libssl-dev libffi-dev python3-dev cargo postgresql postgresql-client gettext curl
```
lijamie98, line 9: Any specific reason to switch from `apt` to `apt-get`?

lijamie98 (approved):

Dockerfile
JakeUrban: I switched from `apk` which is the package manager for alpine linux to `apt-get` which is used for debian systems.

PR #563, opened 2021-11-16
SEP-24: add timezone detection

### LINKED ISSUE #519: Detect user timezone in templates and persist in client app session
Django provides the [TIME_ZONE](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-TIME_ZONE) as a default when datetimes are stored in the database in UTC, but it doesn't detect the users timezone in the case where it doesn't match the default.

Polaris can detect the user's timezone using JavaScript's `Date.getTimezoneOffset()` and pass this to the server so it can be persisted in the user's session cookie. Note that this won't work when `LOCAL_MODE` is `True` since cookies are only used over HTTPS for security reasons.

JakeUrban:
resolves #519


polaris/polaris/middleware.py
```diff
@@ -0,0 +1,39 @@
+import pytz
+from django.utils import timezone
+
+
+class TimezoneMiddleware:
```
lijamie98, line 5: LGTM.

polaris/polaris/sep24/tzinfo.py
```diff
@@ -0,0 +1,37 @@
...
+@api_view(["POST"])
+@parser_classes([JSONParser])
+@renderer_classes([JSONRenderer])
+def post_tzinfo(request: Request) -> Response:
+    if not (
+        request.data.get("sessionId") and request.data.get("sessionOffset") is not None
+    ):
+        return render_error_response("missing required parameters")
+    now = datetime.now(timezone.utc)
+    offset = timedelta(minutes=request.data["sessionOffset"])
```
lijamie98, line 26: `request.data["sessionOffset"]` is inconsistent with line:22:`request.data.get("sessionOffset")`

polaris/polaris/sep24/tzinfo.py
```diff
@@ -0,0 +1,37 @@
...
+    zone = None
+    for tz in map(pytz.timezone, pytz.all_timezones_set):
+        if now.astimezone(tz).utcoffset() == offset:
+            zone = tz.zone
+            break
+    if not zone:
+        return render_error_response("no timezones matched with offset")
+    session = SessionStore(session_key=request.data["sessionId"])
```
lijamie98, line 34: Same as line 26.

lijamie98 (approved):
approved with minor comments.

polaris/polaris/sep24/tzinfo.py
JakeUrban: Thats because the first time we check if the `sessionOffset` variable is present in `request.data`. If it didn't, then `request.data["sessionOffset"]`would throw an exception, whereas `.get("sessionOffset")` would return `None`.

PR #568, opened 2021-11-22
Update reference server's Pipfile.lock after stellar-sdk upgrade

JakeUrban:
The reference server's requirements file was not updated when we updated Polaris' stellar-sdk version.


marcelosalloum (approved):
LGTM!

PR #584, opened 2022-01-12
upgrade pipenv-to-requirements dev package

JakeUrban:
Resolves a dependabot report that the `pipenv` package needs to be upgraded.

I ran `pipenv update pipenv-to-requirements`, which is the package that installs `pipenv` as a development requirement.


lijamie98 (approved):
LGTM

PR #594, opened 2022-02-19
Move all transaction submissions to submit_transaction_task

JakeUrban:
This is a continuation of the work @stfung77 did to upgrade our transaction submission logic and handling.

It adds multi-signature distribution account support back, removes the complex `get_or_create_destination_account()` function, and updates the `submit_transaction_task()` function to handle transactions for creating accounts.

Notable changes to the data model:

* makes `queued_at` nullable. It doesn't makes sense for it to have a value if the `queue` value is `None`.
* Adds a `pending_funding` value to the `Transaction.SUBMISSION_STATUS` enum. This handles the case where the anchor cannot or will not create accounts for end users.

Some things still outstanding:

* test coverage. As of now its at 83%. I'd like to get this back up to 87% or better before merging.
* logging. The new code doesn't have much at this point.
* locking destination accounts when checking for account existence, and if it doesn't exist, creating the account. Currently, two deposit transactions to the same nonexistent account may result in Polaris attempting to create the account twice.


stfung77:
bug fixes, added logging 2608a0b

fixed:
- if the account did not exist already, the transaction would never get fully processed because check_accounts did not queue the transaction for submission (which creates the account)
- race condition - sometimes get_unblocked_transactions could "queue for submission" transactions that have just been created but are missing a trustline (transaction gets queues before the pending_trust status gets set)

stfung77:
changes in ce2af02:

- fix multisig edge case where two or more deposits for the same account generates two+ 'create_account' envelope_xdr's

- discovered and fixed race condition when processing 5+ deposits for a multisig distribution account.
   - deposits can get into a deadlock state where its status is 'pending_trust' but its submission_status is 'blocked' and therefore never gets checked by the "pending trust" task.

PR #606, opened 2022-04-04
bug fixes: don't require SEP-12 'account' request parameter and always pass 'memo' as a string

JakeUrban:
Currently, Polaris requires the `account` SEP-12 request parameter to match account authenticated via SEP-10. However, the SEP states that the `account` request parameter doesn't need to be specified since it is redundant.

This PR updates the request validation logic to only check for a match between account values if the `account` request parameter is provided.

It also ensures that the `memo` value passed to SEP-12 integration functions is always a string. Currently, Polaris will pass `memo` as an integer if `memo_type` is `id`.


stfung77 (approved):

stfung77:
LGTM

PR #611, opened 2022-04-05
Allow url trailing slashes

JakeUrban:
Ensures that all of Polaris' endpoints support requests including and excluding a trailing slash. This issue was identified in stellar/stellar-protocol#1167.

For example, `/auth/?account=XXX` and `/auth?account=XXX` should both be received by the server and handled appropriately. Currently, Polaris deployments will only properly handling requests that do not include a trailing slash.

To do this, all URL paths in Polaris must be converted to regular expressions, where a trailing slash is added as an optional character.

In an unrelated change, this PR also fixes a potential `TypeError` in the example server.

### DIFF
--- a/example/server/integrations/sep31.py
+++ b/example/server/integrations/sep31.py
@@ -75,11 +75,11 @@ class MySEP31ReceiverIntegration(SEP31ReceiverIntegration):
             receiving_user.save()
 
         transaction.amount_fee = round(
-            transaction.asset.send_fee_fixed
-            or 0
+            (transaction.asset.send_fee_fixed or 0)
             + (
-                transaction.asset.send_fee_percent
-                or 0 / Decimal(100) * transaction.amount_in
+                (transaction.asset.send_fee_percentor or 0)
+                / Decimal(100)
+                * transaction.amount_in
             ),
             transaction.asset.significant_decimals,
         )
--- a/example/server/integrations/sep31.py
+++ b/example/server/integrations/sep31.py
@@ -76,10 +76,10 @@ class MySEP31ReceiverIntegration(SEP31ReceiverIntegration):
 
         transaction.amount_fee = round(
             transaction.asset.send_fee_fixed
+            or 0
             + (
                 transaction.asset.send_fee_percent
-                / Decimal(100)
-                * transaction.amount_in
+                or 0 / Decimal(100) * transaction.amount_in
             ),
             transaction.asset.significant_decimals,
         )
--- a/example/server/integrations/sep31.py
+++ b/example/server/integrations/sep31.py
@@ -77,7 +77,7 @@ class MySEP31ReceiverIntegration(SEP31ReceiverIntegration):
         transaction.amount_fee = round(
             (transaction.asset.send_fee_fixed or 0)
             + (
-                (transaction.asset.send_fee_percentor or 0)
+                (transaction.asset.send_fee_percent or 0)
                 / Decimal(100)
                 * transaction.amount_in
             ),
--- a/polaris/polaris/sep1/urls.py
+++ b/polaris/polaris/sep1/urls.py
@@ -1,5 +1,6 @@
 """This module defines the URL patterns for the `/.well-known/stellar.toml` endpoint."""
-from django.urls import path
+from django.urls import re_path
+
 from polaris.sep1.views import generate_toml
 
-urlpatterns = [path("stellar.toml", generate_toml)]
+urlpatterns = [re_path(r"^stellar.toml/?$", generate_toml)]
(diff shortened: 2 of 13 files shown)


example/server/integrations/sep31.py
```diff
@@ -76,10 +76,10 @@ def process_post_request(
 
         transaction.amount_fee = round(
             transaction.asset.send_fee_fixed
+            or 0
```
leighmcculloch, line 79: I don't think this does what you want. In Python `+` has higher precedence than `or`, so the `0 + ...` would occur together, but that doesn't matter, since the `or` early exits and only `.send_fee_fixed` is returned if it is set.

I think you want:
```python
(transaction.asset.send_fee_fixed or 0) + ...
```

Ref:https://docs.python.org/3/reference/expressions.html#operator-precedence

polaris/polaris/sep1/urls.py
```diff
@@ -1,5 +1,6 @@
 """This module defines the URL patterns for the `/.well-known/stellar.toml` endpoint."""
-from django.urls import path
+from django.urls import re_path
+
 from polaris.sep1.views import generate_toml
 
-urlpatterns = [path("stellar.toml", generate_toml)]
+urlpatterns = [re_path(r"^stellar.toml/?$", generate_toml)]
```
leighmcculloch, line 6: Is there a way to implement this without re_path so that a large number of regular expressions aren't being evaluated on every request? Regular expressions evaluated on user input can be problematic ([ref](https://rules.sonarsource.com/python/RSPEC-4784)). They're also generally very costly, so running a bunch of them on an incoming http request to try and match to a single route may degrade Polaris' performance.

Maybe `APPEND_SLASH=True` ([ref](https://docs.djangoproject.com/en/dev/ref/settings/#append-slash)) would address some of the issue? It is currently set to `False` by the looks of it:
https://github.com/stellar/django-polaris/blob/07c31b4f0983073f7c4866f1c3c070863c8d8f25/polaris/settings.py#L71

Would it also work to write a middleware that automatically matches routes

JakeUrban: `APPEND_SLASH` works by returning a 301 redirect to the URL including the slash.

Ex. `/auth?account=` --> `/auth/?account=`

This works for GET requests, but not for POST (or any method that uses a request body), since request bodies are not included in 301 redirects. (see the third comment on [this SO answer](https://stackoverflow.com/questions/1596552/django-urls-without-a-trailing-slash-do-not-redirect)). I've reproduced this issue too.

However, I think there is a good middle ground we can take.

For endpoints that only support GET (which is the majority of Polaris endpoints), we can use the normal `path()` method vs. `re_path()` and ensure `APPEND_SLASH = True`. Then, we can use `re_path()` for endpoints that support the POST, PUT, or DELETE methods.

The `APPEND_SLASH` functionality on kicks in when a request doesn't match any of the URLs defined ([ref](https://docs.djangoproject.com/en/4.0/ref/settings/#append-slash)), so it won't attempt to redirect any POST/PUT/DELETE requests that match the patterns defined in the `re_path()` calls.

leighmcculloch: >Or, could a middleware help match to routes with/without slash by normalizing the incoming request paths?

What about this ☝🏻? This should be a relatively small amount of code I think.

JakeUrban: That would work, but it is an additional setup/configuration step for users of Polaris. Ideally this functionality comes out of the box.

JakeUrban: Oh no... turns out my suggested approach doesn't work because auth headers are also not persisted when requests are redirected.

JakeUrban: Is normalizing request paths possible? I would think request paths are immutable, thats why redirects are required right?

leighmcculloch: I don't know, sorry.

JakeUrban: I did a bit of digging into how Django resolves URLs using `path()` and `re_path()`. Turns out Django uses Python's `re` regular expression package for matching all URL paths, regardless of whether `path()` or `re_path()` is used ([source](https://github.com/django/django/blob/baf9604ed8fed3e6e7ddfaca2d83c377c81399ae/django/urls/conf.py#L94-L95)). 

To address your points about security and performance, none of the inefficient / sensitive patterns mentioned in [RSPEC-4784](https://rules.sonarsource.com/python/RSPEC-4784) are used in our URL paths. Additionally, this rule is deprecated in favor of [RSPEC-2631](https://rules.sonarsource.com/python/RSPEC-2631), which states that user input should not be used to as regex patterns.

Given these points, I think I'd like to move forward with the current changes in the PR. I don't think any of our patterns can be leveraged to execute denial of service attacks.

leighmcculloch (approved):
🚀

PR #623, opened 2022-05-06
refactor project structure for ease of development

JakeUrban:
This PR significantly changes the code structure of the project in order to make contributing to the project easier.

The primary benefit is that contributors can now run the reference server using the adjacent `polaris` source code without docker. Contributors should be able to clone the repository, use the `migrate` and `collectstatic` commands to initialize the application, and `runserver` to stand up the reference server.

The second benefit is moving away from the legacy build system `setuptools` and using pipenv for dependency management to the new and unified python dependency/build system: poetry.

The `Dockerfile` and `docker-compose.yml` files have been updated to support the new structure, so using docker is still supported.

Running tests is also easier. You can now run `pytest` without the `-c polaris/pytest.ini` argument.

There are a few unrelated changes added in this PR as well, although I'm happy to move to a separate PR:

- Swapped the old Stellar rocket icon with the current logo
- Slightly adjusted some padding on the base.css stylesheet for the SEP-24 interactive flow
- The default fee calculation function will no longer raise a `ValueError` if there are no `fee_fixed` or `fee_percent` values configured.


yuriescl:
+1 for Poetry

stfung77 (approved):
this looks great! i'll try running it from source tomorrow

stfung77:
just tried this on windows, definitely a lot easier to set up!

PR #630, opened 2022-05-12
move dev-server extras to dependencies

JakeUrban:


poetry.lock
```diff
@@ -302,14 +302,14 @@ bcrypt = ["bcrypt"]
...
-version = "3.11.0"
+version = "3.12.0"
 description = "django-cors-headers is a Django application for handling the server headers required for Cross-Origin Resource Sharing (CORS)."
 category = "main"
 optional = false
 python-versions = ">=3.7"
 
 [package.dependencies]
-Django = ">=2.2"
+Django = ">=3.2"
```
marcelosalloum, line 312: This seems like a big update 👀 .

stfung77 (approved):

marcelosalloum (approved):

poetry.lock
JakeUrban: Thanks haha, it would be but the 2 was a typo. Polaris has been on Django 3 for awhile.

PR #638, opened 2022-06-09
bug fix: add network passphrase param to `get_balance_id()`

### LINKED ISSUE #637: Network passphrase is necessary
https://github.com/stellar/django-polaris/blob/73f4bbe7f874f67ad3f6d6476c64feb2f6dd3c83/polaris/management/commands/process_pending_deposits.py#L711

This line fails because network_passphrase if necessary arg for TransactionEnvelope.from_xdr method

JakeUrban:
resolves #637 

Currently, Polaris cannot decode the envelope of a deposit transaction it submits to the network if it uses claimable balances. This PR fixes this bug.

Thanks @nikita-gorodeckij for reporting.


JakeUrban:
In the interest of time, I am not adding tests in this PR, but I have confirmed that this change fixes the issue. Tests should certainly be added after the release.

stfung77 (approved):

PR #656, opened 2022-08-17
SEP-24: make `account` param optional in `POST /deposit`

JakeUrban:
We made the `account` parameter optional in SEP-24's `POST /deposit` endpoint since it can default to the authenticated account. This PR updates Polaris with that change.


stfung77 (approved):
LGTM

PR #661, opened 2022-11-18
Add Circle USDC support

stfung77:
What?
Add Circle USDC support to polaris forms
Add cli functionality to testnet.py to add/delete assets and add Circle USDC to distribution account (for testnet resets)

Why?
Wallet SDK team would like to test with a SEP-24 anchor that doesn’t issue an asset


polaris/management/commands/testnet.py
```diff
@@ -320,3 +417,162 @@ def get_balance(self, code, issuer_public_key, json) -> Optional[str]:
                 and balance_obj.get("asset_issuer") == issuer_public_key
             ):
                 return balance_obj["balance"]
+
+    def create_trustline(self, code :str, asset_issuer_public_key: str, distribution: Keypair):
```
JakeUrban, line 421:
```suggestion
    def create_trustline(self, code: str, asset_issuer_public_key: str, distribution: Keypair):
```

It may be helpful to run `black` on this code to format it

polaris/management/commands/testnet.py
```diff
@@ -320,3 +417,162 @@ def get_balance(self, code, issuer_public_key, json) -> Optional[str]:
...
+        envelope.sign(distribution)
+        try:
+            self.server.submit_transaction(envelope)
+        except BaseHorizonError as e:
+            print(
+                f"Failed to create trustline "
+                f"Result codes: {e.extras.get('result_codes')}"
+            )
+        else:
+            print(f"Success! Trustline to {code}:{asset_issuer_public_key} created!")
```
JakeUrban, line 448: We have an `add_balance()` function, can we refactor this and that function so we're not duplicating code?

polaris/management/commands/testnet.py
```diff
@@ -320,3 +417,162 @@ def get_balance(self, code, issuer_public_key, json) -> Optional[str]:
...
+        
+        print(f"\nAdding asset {asset_code}:{issuer} to database")
+        Asset.objects.create(
+            code=asset_code,
+            issuer=issuer,
+            distribution_seed=distribution_seed,
+            sep24_enabled=options.get("sep24_enabled") or True,
+            deposit_enabled=options.get("deposit_enabled") or True,
+            withdrawal_enabled=options.get("withdrawal_enabled") or True,
+            symbol="$"
```
JakeUrban, line 473: I think we should use whichever default value is defined in the DB, which I think is `False`. If we're trying to create a generic add-asset CLI command, we also. need to provide the ability to specify all fields, including symbol.

polaris/management/commands/testnet.py
```diff
@@ -320,3 +417,162 @@ def get_balance(self, code, issuer_public_key, json) -> Optional[str]:
...
+
+        response = requests.get(url, headers=headers)
+        wallets = response.json().get("data")
+        max_wallet_balance = ("", float(0))  
+        for wallet in wallets:
+            for balance in wallet["balances"]:
+                if balance["currency"] == "USD" and float(balance["amount"]) > max_wallet_balance[1]:
+                    max_wallet_balance = (wallet["walletId"], float(balance["amount"]))
+        print(f"\nLargest circle wallet: {max_wallet_balance}")
+        return max_wallet_balance
```
JakeUrban, line 511: I don't think we should have this implicit rule where the wallet with the largest balance is used. Instead we should require the dev to pass the wallet ID. If we don't want to do that, we should default to the master wallet ID and have the dev confirm before we use it.

server/forms.py
```diff
@@ -81,6 +81,10 @@ class SelectAssetForm(forms.Form):
                 "stellar:SRT:GCDNJUBQSX7AJWLJACMJ7I4BC3Z47BQUTMHEICZLE6MU4KQBRYG5JY6B",
                 "Stellar Reference Token (SRT)",
             ),
+            (
+                "stellar:USDC:GBD6ZRRI4FZW55G7NBRMF2NY2DANKHXY63J4KKIQT5GWNWEUZWZ6MS2A",
```
JakeUrban, line 85: This is not Circle's testnet address, can we use the real testnet USDC?

JakeUrban: `GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5`

JakeUrban: We obviously wouldn't be able to issue USDC, but we can hold a balance and get USDC from their faucet.

JakeUrban:
Also, have you tested the SEP-24 deposit & withdraw flow with USDC? I thought there would be more changes necessary than just adding an entry in the menu.

polaris/management/commands/testnet.py
JakeUrban: Personally I don't think its worth the effort. Using `python manage.py shell` and adding the asset using the Django model is not terribly hard, especially with good example documentation and description of the fields.

JakeUrban: I like your middle-ground approach though, allowing devs to create the asset with the enabled SEPs but not specify fees or min/max limits.

server/static/polaris/stellar.toml
```diff
@@ -16,6 +16,13 @@ status = "test"
 is_asset_anchored = false
 desc = "Stellar Reference Token (SRT) is an asset issued on testnet and is used as an anchored asset for this reference server for demonstration and testing purposes."
 
+[[CURRENCIES]]
+code = "USDC"
+issuer = "GBD6ZRRI4FZW55G7NBRMF2NY2DANKHXY63J4KKIQT5GWNWEUZWZ6MS2A"
```
JakeUrban, line 21: Wrong issuer :-)

JakeUrban (approved):
Nicely done, thanks. The only thing that would be included in the next release is the testnet CLI changes, but since we don't have a future release planned we should probably just release it now. We can sync later to go over how thats done.

PR #680, opened 2023-03-01
Added the mIssing OPERATION_SEND constant into polaris settings.py file

santoshchaurasiya:


JakeUrban (approved):
Thanks for adding! While reviewing I also noticed we were missing the `withdrawal-exchange` case from `Transaction.KIND`, so I added that as well.

PR #687, opened 2023-04-12
Added body_scripts block to base template

yuriescl:
There are some cases where the Anchor wants to override Polaris's `<script>` imports, for example, I'm trying to override the feeTable JS code and import my custom JS before the remaining `<script>` Polaris imports, but right now I can't do it using the block `extra_body` since it's loaded before the scripts. My only choice left was to create a duplicate `base.html` in my project but that breaks all my `extend` template hierarchy.


yuriescl:
@JakeUrban this PR appears to have been missed from latest release. I think it would be a useful addition to the software.

JakeUrban:
Sorry @yuriescl you're right. This change seems straightforward so I'll approve, but can you update your branch so that it includes the github action for running tests? It won't let me merge the PR until they've passed.

yuriescl:
@JakeUrban done

JakeUrban (approved):

PR #692, opened 2023-08-20
Missing SEP 9 Fields

kaning:
Some fields from the protocol were missing in the list of authorised SEP 9 fields. 
These are: 
- 'sex' 
- 'proof_of_liveness'
-  'proof_of_income'

This cause GET /kyc/customer to fail where the and external (or internal) KYC service had made provisions for all fields listed [here](https://github.com/stellar/stellar-protocol/blob/master/ecosystem/sep-0009.md#natural-person-fields). 

This PR just adds the missing natural person fields to the list of Authorized fields so that this [check](https://github.com/mykobo/django-polaris/blob/master/polaris/sep12/customer.py#L338) passes.


JakeUrban (approved):
Thanks for adding @kaning! I'll merge and publish a release.

PR #699, opened 2023-08-21
Fix poetry installation

JakeUrban:
Installs poetry on ubuntu without actions/setup-python@v4

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,11 @@ jobs:
       - name: Install Python
         run: apt-get update && apt-get install -y python3-dev python3-venv
       - name: Install Poetry
-        run: export POETRY_HOME=/opt/poetry && python3 -m venv $POETRY_HOME && $POETRY_HOME/bin/pip install poetry==1.6 && ln -s $POETRY_HOME/bin/poetry /usr/bin/poetry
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -13,11 +13,17 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: '3.11' 
-          cache: poetry
+      - name: Install Python
+        run: apt-get update && apt-get install -y python3-dev python3-venv
```
Ifropc, line 17: Can we use python3.10-dev here and everywhere else to avoid potential breaking on new python version being released?

.github/workflows/release.yml
```diff
@@ -13,11 +13,17 @@ jobs:
...
+      - name: Install Poetry
+        run: |
+          apt-get update
+          apt-get install -y python3-dev python3-venv
+          mkdir -p /usr/local/poetry
+          python3 -m venv /usr/local/poetry
```
Ifropc, line 23: same here, suggesting using python3.10 instead

Ifropc (approved):

PR #724, opened 2025-02-12
chore: add support for Django 4.2+, and drop support for Django 3.x.

overcat:
See https://endoflife.date/django

A few days ago, I received a request from a team asking to make polaris support Django 4.2, so I created this PR.

This PR should be compatible with any version above 4.2; however, as a compromise with the package manager, I removed support for Python 3.9 because Django 5 requires Python 3.10+.

Additionally, I am looking to see if there are any package managers that can work well with tox so that we can test the minimum supported dependencies and the latest dependencies separately.


settings.py
```diff
@@ -91,7 +91,6 @@
 LANGUAGE_CODE = "en-us"
 TIME_ZONE = "America/Los_Angeles"
 USE_I18N = True
-USE_L10N = True
```
overcat, line 94: https://github.com/django/django/pull/14846/files#diff-b6f3108584c2db0b4c53253b6b3b5d5c60fbb3a82e7ed6544922fe689631fa43R24-R29

JakeUrban (approved):
This LGTM, thanks for making the change.

